PR #17, opened 2015-02-27
Adding placeholder functionality to object picker field

nodoherty:
So values can be defined in an i18n file


DominicWatson:
Needs to be targeted at our develop branch.

PR #19, opened 2015-03-10
Adding additional functions for accessing member benefits details

nodoherty:


system/services/websiteUsers/WebsiteLoginService.cfc
```diff
@@ -279,6 +279,30 @@ component output=false singleton=true autodoc=true displayName="Website login se
...
+              id           = getLoggedInUserId()
+            , selectFields = [ "benefits.id" ]
+            , forceJoins   = "inner"
+        );
+
+        return ValueArray( benefits.id );
+    }
+
+    /**
+	 * Returns true / false depending on whether or not a user has access to a specific benefit
```
DominicWatson: White space here is dodgy, plus could have better comments:

> Returns true / false depending on whether or not a user has access to any of the supplied benefits
>
> @benefits.hint Array of benefit IDs. If the logged in user has any of these benefits, the method will return true

system/services/websiteUsers/WebsiteLoginService.cfc
```diff
@@ -279,6 +279,30 @@ component output=false singleton=true autodoc=true displayName="Website login se
 		return _getSessionService().deleteVar( "websitePostLoginUrl" );
 	}
 
+	/**
+	 * Gets an array of benefits associated with the logged in user
```
DominicWatson: I think this is more accurate:

> Gets an array of benefit IDs associated with the logged in user

PR #25, opened 2015-05-21
Release 10.1.0

john5onCheng:
pull request :)


DominicWatson:
This isn't going to work I'm afraid! What you need to do is do your work in its own branch. Then submit the PR. There should be no direct commits to any release branch but also all Pull requests should be worked on in their own branch.

john5onCheng:
sorry i got it wrong

PR #28, opened 2015-05-28
PRESIDE-24

john5onCheng:
Add i18n for new filetype


DominicWatson:
Could you add an entry in `Config.cfc` for the file type and commit that?

john5onCheng:
Hi Dom, yes we already did that in previous PR

PR #30, opened 2015-06-09
PRESIDECMS-51

john5onCheng:
Show all page type in trash and skip maxDepth if _heirarchi_depth is null value


DominicWatson:
Thanks @john5onCheng :)

Closing now as merged manually.

PR #32, opened 2015-06-12
PRESIDECMS-67

john5onCheng:
Remove the additional header for widget dialog


DominicWatson:
Nice one JC. Could you check the produced markup to see a double `<div class="page-content">` div is now created? I suspect you can remove that div from the view too.

PR #34, opened 2015-06-18
PRESIDECMS-79

josephpbuarao:
Create Datetime picker control


DominicWatson:
Thanks for this @josephpbuarao. However, I tried it and it is not quite ready for core. Could we try implementing this instead: https://tarruda.github.io/bootstrap-datetimepicker/

The benefit here is that this is based on a single plain text field for the resulting datetime value + that makes it play nicely with our validation plugins, etc. + less to go wrong.

We'd want it configured without seconds :). Finally, the default value for the picker should be an empty string and **not** the current date.

D

PR #44, opened 2015-07-31
PRESIDECMS-130

john5onCheng:
Add new maxfilesize attribute to asset picker property


DominicWatson:
Thanks @john5onCheng :)

PR #45, opened 2015-08-09
PRESIDE-131 add link to assets on images

seakchiew:
added asset picker on insert image form so that user can select link to asset on images


DominicWatson:
I wonder if a cleaner way to implement this would be to have a 'link picker' rather than plain text input for the link. That will cover assets, pages plus anything else we later implement.

Problem with that of course is backward compatibility so the renderer would need to check the format of the link - i.e. if it is a UUID, treat it as a preside link, otherwise, treat it as a plain text link.

What do you reckon?

DominicWatson:
Scratch that ^^^^. Instead, let's add a page picker while we're here and improve the labelling to make it as clear as possible (as discussed). So:
- `system/forms/richeditor/image.xml` - add field for site tree page picker (use `sitetreePagePicker` control). Also add placeholder for the current link field
- `system/forms/i18n/cms.properties` - change link field label to be 'HREF' and add a place holder like 'e.g. http://www.google.com'. Also add labels for the asset and page fields
- `system/views/renderers/asset/image/richEditor.cfm` - add logic to detect if a page was selected and link to that

PR #48, opened 2015-09-10
PRESIDECMS-151 Return result within interceptionArgs struct

vivienwong:
Return result within interceptionArgs


DominicWatson:
Thanks @vivienwong :+1: :)

PR #53, opened 2015-10-22
Added - Core widget: iframe

josephpbuarao:


system/i18n/widgets/iframewidget.properties
```diff
@@ -0,0 +1,5 @@
+title=Iframe Widget
+description=Run HTML Code on the site
```
DominicWatson, line 2: Title should be "Iframe"
Description should be "Configure and embed an Iframe in your content"

DominicWatson:
I'm not sure the approach here is quite right @josephpbuarao. I'd expect to be able to configure:

URL for the iframe
Height
Width
Border (yes/no - default to no)

There should not be a title and the renderer should then take care of rendering the iframe html for you (you shouldn't have to enter it yourself).

We should also create a custom placeholder for the widget for all our core widgets. I can point you to a good example to copy from :)

DominicWatson:
Closing as duplicate

PR #55, opened 2015-10-27
PRESIDECMS-115

cfmitrah:
Strip file extension from file name when downloading from asset manager


DominicWatson:
This is being merged manually. Thanks @cfmitrah!

PR #56, opened 2015-10-27
mainNavigation changes inspired by MuraCMS dspPrimaryNav function  

cfmitrah:
mainNavigation changes inspired by MuraCMS dspPrimaryNav function  
https://gist.github.com/ronnieduke/87c31b7c7f733674bcf9#file-template-variables-L52


DominicWatson:
Nice one - I tidied it up a little, renamed `kids` to `children` and set the defaults so that the resulting default markup has not changed from what we have now.

This has been merged into the upcoming release branch. Thanks.

PR #59, opened 2015-10-30
Added - Iframe Widget

josephpbuarao:
Hi @DominicWatson 

Good day! I've amended the iframe core widget, kindly check it and let me know your feedback.

Thanks and have a great day ahead.

### DIFF
--- a/system/forms/widgets/iframewidget.xml
+++ b/system/forms/widgets/iframewidget.xml
@@ -1,11 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<form>
+<form i18nBaseUri="widgets.iframewidget:">
 	<tab>
 		<fieldset>
-			<field name="url" 	   label="widgets.iframewidget:form.url.label" required="true" />
-			<field name="border"   label="widgets.iframewidget:form.border.label" control="yesNoSwitch" />
-			<field sortorder="60" name="width"    control="spinner" required="false" label="widgets.iframewidget:form.width.label" default="560" />
-			<field sortorder="60" name="height"    control="spinner" required="false" label="widgets.iframewidget:form.height.label" default="315" />
+			<field name="url"     required="true" />
+			<field name="border"  control="yesNoSwitch" />
+			<field sortorder="60" name="width"    control="spinner" required="false" default="560" />
+			<field sortorder="60" name="height"   control="spinner" required="false" default="315" />
 		</fieldset>
 	</tab>
 </form>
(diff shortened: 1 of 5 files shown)


system/views/widgets/iframewidget.cfm
```diff
@@ -0,0 +1,14 @@
+<cfparam name="args.url" 	default="" />
+<cfparam name="args.width"  default="560" />
+<cfparam name="args.height" default="315" />
+<cfparam name="args.border" />
+
+<cfif args.border EQ true >
```
DominicWatson: Should be `IsTrue( args.border )` (this checks that the value is truthy - it is a Preside helper function)

system/views/widgets/iframewidget.cfm
```diff
@@ -0,0 +1,14 @@
+<cfparam name="args.url" 	default="" />
```
DominicWatson: Very picky, but could these 'default' attributes line up (use spaces for alignment and not tabs)

system/forms/widgets/iframewidget.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<form>
```
DominicWatson: We should use the newish `i18nBaseUri` attribute on the `form` element here for all the labelling. So `<form i18nBaseUri="widgets.iframewidget:">`. You can then remove all the label attributes from the fields and use the following convention in the `.properties` file:

```
field.url.title=URL
field.url.placeholder=e.g. http://myiframesite.com/iframe.html
field.url.help=Some help text about URL
```

etc.

DominicWatson:
Looking good, I've made a few comments above that should be addressed. We should also use a custom placeholder for the widget. You can do this simply by creating a view at `/views/widgets/iframewidget/placeholder.cfm` - it should contain something like:

```
<cfparam name="args.url" default="" />
<cfoutput>#translateResource( 'widgets.iframewidget:title' )#: #args.url#</cfoutput>
```

system/forms/widgets/iframewidget.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<form i18nBaseUri="widgets.iframewidget:">
+	<tab>
+		<fieldset>
+			<field name="url"     required="true" />
+			<field name="border"  control="yesNoSwitch" />
+			<field sortorder="60" name="width"    control="spinner" required="false" default="560" />
```
DominicWatson: Sorry, missed this before. Could you sort out the sort order on these? (width and height both have sortorder=60 - all fields should have sort order here, 10, 20, 30, 40). Also, once that's sorted, can make sure the attributes vertically align.

DominicWatson:
Looking good :) this has been manually merged.

PR #61, opened 2015-11-03
Added - Javascript Snippet Core Widget

josephpbuarao:
Hi @DominicWatson ,

Kindly check the JavaScript snippet widget when you have chance.

Thank you!

### DIFF
--- a/system/forms/widgets/jssnippetwidget.xml
+++ b/system/forms/widgets/jssnippetwidget.xml
@@ -1,11 +1,14 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <form i18nBaseUri="widgets.jssnippetwidget:">
 	<tab>
-		<fieldset>
-			<field sortorder="10" name="source_url"      required="true" />
+		<fieldset id="externaljs">
+			<field sortorder="10" name="source_url"      control="textinput"/>
 			<field sortorder="20" name="async"           control="yesNoSwitch" />
 			<field sortorder="30" name="defer"           control="yesNoSwitch" />
-			<field sortorder="40" name="internal_script" control="textarea" />
+		</fieldset>
+
+		<fieldset id="inlinejs">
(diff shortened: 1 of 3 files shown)


system/i18n/widgets/jssnippetwidget.properties
```diff
@@ -0,0 +1,17 @@
+title=Javascript Snippet
+description=Configure and embed an Iframe in your content
```
DominicWatson: Still says iframe from copy and paste ;)

system/forms/widgets/jssnippetwidget.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<form i18nBaseUri="widgets.jssnippetwidget:">
+	<tab>
+		<fieldset>
+			<field sortorder="10" name="source_url"      required="true" />
```
DominicWatson: The Source_URL cannot be required if we optionally have internal_script. Also, would be good to separate these out into fieldsets with titles "External javascript" + "Inline javascript". All fields to go in the "External javascript" fieldset except for "internal_script" (which should be renamed to "inline_script").

system/views/widgets/jssnippetwidget/index.cfm
```diff
@@ -0,0 +1,24 @@
...
+			<cfset method = "defer" />
+		<cfelse>
+			<cfset method = "" />
+		</cfif>
+
+		<script type="text/javascript" src="#args.source_url#" #method#></script>
+	</cfif>
+
+	<cfif !IsEmpty( args.internal_script )>
+		#args.internal_script#
```
DominicWatson: This should be surrounded by `<script type="text/javascript"></script>` tags. The idea being that the user can enter just the raw JS.

DominicWatson:
Manually merged

PR #64, opened 2015-11-10
PRESIDE-217

john5onCheng:
Update asset dimension when changing version


DominicWatson:
Good spot and fix @john5onCheng, thank you - merged into upcoming 10.3.0 release :)

PR #70, opened 2015-11-30
show more info about latest release & available versions

cfmitrah:


DominicWatson:
This is great thanks @cfmitrah :) Have manually merged this into the upcoming 10.4.0 release. This PR should auto close when that release is officially released sometime this month (before 25th December).

Thanks again

PR #71, opened 2015-12-01
Added - Image derivative setting is now available on the preside editor

josephpbuarao:
Hi @DominicWatson 

Good day! Please review my pull request if you have time soon. 

Thank you very much! :)


box.json
```diff
@@ -16,7 +16,9 @@
     "description":"",
```
DominicWatson, line 16: Changes to this file should be reverted I think - has nothing to do with image derivatives :)

system/config/Config.cfc
```diff
@@ -225,6 +226,9 @@ component output=false {
 			, toolbars    = _getCkEditorToolbarConfig()
 		};
 
+		// settings.ckeditor.linkStyles.append(  "button" );
```
DominicWatson: Please remove these commented out lines.

system/config/Config.cfc
```diff
@@ -401,6 +405,7 @@ component output=false {
 
 		derivatives.adminthumbnail = {
 			  permissions = "inherit"
+			, inEditor    = true
```
DominicWatson: I think the admin derivatives should not be available in the picker by default.

system/handlers/formcontrols/DerivativePicker.cfc
```diff
@@ -0,0 +1,34 @@
...
+		var args.values = [];
+		var args.width  = [];
+		var args.height = [];
+
+		for( var derivative in derivatives.value ) {
+		    args.values.append( derivative );
+			args.labels.append( translateResource( uri="derivatives:#derivative#.title", defaultValue="derivatives:#derivative#.title" ) );
+		}
+
+		for( var transformation in derivatives.dimension ) {
```
DominicWatson: This is not quite right - a derivative can be more than just a resize - it could potentially add a watermark and do other things. Instead of having the dimensions inserted into the image URL, we need the name of the derivative and nothing more. (this note applies to other areas of the code).

system/i18n/cms.properties
```diff
@@ -560,6 +560,7 @@ ckeditor.imagepicker.spacing_right.label=Right (px)
 ckeditor.imagepicker.spacing_bottom.label=Bottom (px)
 ckeditor.imagepicker.spacing_left.label=Left (px)
 
+ckeditor.imagepicker.image_derivative.label=Image Derivative
```
DominicWatson: I think we should rename this to "Preset". "Image Derivative" is perhaps too technical for end-users.

system/services/assetManager/AssetManagerService.cfc
```diff
@@ -936,6 +936,29 @@ component {
...
 	}
 
+	public any function listEditorDerivates(){
+		var derivatives = _getConfiguredDerivatives();
+		var publicDerivatives = structNew();
+		var value = [];
+		var dimension = [];
+
+		for( var derivative in derivatives ) {
+
```
DominicWatson: The conditional logic below can be expressed more concisely:

```
if ( IsBoolean( derivatives[ derivative ].inEditor ?: "" ) && derivatives[ derivative ].inEditor ) {
    // ...
}
```

system/services/assetManager/AssetManagerService.cfc
```diff
@@ -936,6 +936,29 @@ component {
...
+			if ( derivatives[ derivative ].keyExists( "inEditor" ) ) {
+
+				var inEditor = derivatives[derivative].inEditor ?: false;
+
+				if( inEditor ){
+
+				    publicDerivatives.value     = value.append( derivative );
+				    publicDerivatives.dimension = dimension.append( _getPreconfiguredDerivativeTransformations( derivative ) );
```
DominicWatson: Given the previous note around not wanting dimensions - we should just build an array of "publicDerivatives" so this can just be:

```
publicDerivatives.append( derivative );
```

system/views/formcontrols/selectDerivatives/index.cfm
```diff
@@ -0,0 +1,50 @@
+<cfscript>
```
DominicWatson: I don't think we need this file. After building the arrays of `values` and `labels` in the handler, you should be able to just render the "select" form control. Use extra classes for any javascript that needs to interact with the field. Hardcoding the ID of the field means there can be only one picker per page which should be avoided as an approach.

DominicWatson:
This is looking good @josephpbuarao - I've made a number of comments above; let's catchup asap.

josephpbuarao:
Hi @DominicWatson ,

Good day! Thank you very much for your feedback. I've updated it according to your suggestions. Also, I got one blocker, the drop down select for quality can't be disabled. Could you please advise. 

Thank you very much!

system/handlers/formcontrols/DerivativePicker.cfc
```diff
@@ -0,0 +1,19 @@
+component output=false {
+
+	property name="assetManagerService"  inject="AssetManagerService";
+
+	public string function index( event, rc, prc, args={} ) output=false {
+
+		var derivatives = assetManagerService.listEditorDerivates();
+		args.labels = [];
```
DominicWatson: We need to return an empty string here in the case where there are no derivatives, so that the form field doesn't show. So:

```
if ( !derivatives.len() ) {
    return "";
}
```

system/i18n/derivatives.properties
```diff
@@ -0,0 +1 @@
+none.title=No Derivative Selected
```
DominicWatson: This should be "No preset selected" :)

system/services/assetManager/AssetManagerService.cfc
```diff
@@ -936,6 +936,26 @@ component {
 		return settings;
 	}
 
+	public any function listEditorDerivates(){
+		var derivatives = _getConfiguredDerivatives();
+		var publicDerivatives = ["none"];
```
DominicWatson: The "none" doesn't belong here really. It is specific logic for the form control and not the service logic. If I do `.listEditorDerivatives()` I expect to just get actual valid derivatives.

system/services/assetManager/AssetManagerService.cfc
```diff
@@ -936,6 +936,26 @@ component {
 		return settings;
 	}
 
+	public any function listEditorDerivates(){
```
DominicWatson: Typo - should be `listEditorDerivatives` (you have `listEditorDerivates`)

system/handlers/formcontrols/DerivativePicker.cfc
```diff
@@ -0,0 +1,19 @@
...
+		args.values = [];
+
```
DominicWatson, line 15: The `listEditorDerivatives` method should not include a 'none' entry (see comment further down), so we should manually add the 'none' entry for the form control here (but only if there are any derivatives to show). The entry should have an empty value, so:

```
args.values = [ "" ];
args.labels = [ translateResource( "derivatives:none.title" ) ];
```

system/services/assetManager/AssetManagerService.cfc
```diff
@@ -936,6 +936,26 @@ component {
...
+		var derivatives = _getConfiguredDerivatives();
+		var publicDerivatives = ["none"];
+
+		for( var derivative in derivatives ) {
+
+			if ( derivatives[ derivative ].keyExists( "inEditor" ) ) {
+
```
DominicWatson: There's a lot of extraneous whitespace here. would prefer there not to be all these extra empty lines.

system/services/rendering/ContentRendererService.cfc
```diff
@@ -223,15 +223,27 @@ component singleton=true output="false" {
...
 
 				args.delete( "asset" );
 				args.delete( "placeholder" );
+				args.delete( "derivative" );
 
-				renderedImage = _getAssetRendererService().renderAsset(
-					  assetId = embeddedImage.asset
-					, context = arguments.context
-					, args    = args
+				if( Len( Trim( derivative ) ) && derivative NEQ "none" ){
```
DominicWatson, line 233: After changes, we shouldn't  need the `&& derivative NEQ "none"` condition.

DominicWatson:
Looking good @josephpbuarao, nearly there :) I've made some comments for some tweaks that need making.

josephpbuarao:
Hi @DominicWatson ,

Thank you very much for your feedback. I already implemented the changes you have requested.

Thank you!

system/assets/js/admin/presidecore/formFields.js
```diff
@@ -53,4 +53,28 @@
         sideBySide:true
 	});
 
+
+	var width       = $('.image-dimensions-picker-width');
```
DominicWatson: This is not quite working for me - in the JS here we have two issues:
1. Using element IDs instead of classes (#derivative_chosen, #derivative) - there could be multiple instances of this in a page potentially - the JS should work for each form independently
2. Refering to classes outside of the context of an individual derivative form instance

ultimately, we need this:

```
for each derivative field:
    find its related height, width and quality form fields
    set disabled/enabled status of those fields according to value of derivative field
    when derivative field changes, recalculate disabled/enabled status of those fields
end for;
```

For disabling the quality drop down: try something like `$qualityInputField.prop( "disabled", true ).data( 'uberSelect' ).search_field_disabled()`.

josephpbuarao:
Hi @DominicWatson ,

Good day! Please see the revised formfields.js if you have a chance.

Thank you and have a great day ahead.

DominicWatson:
W00t, tested this and all working well :). Has been merged into upcoming 10.4.0 release. Good work @josephpbuarao

PR #74, opened 2015-12-02
PRESIDECMS-231 :: Domain and Path next to slug field

cfmitrah:
Domain and Path next to slug field


system/assets/js/admin/presidecore/formFields.js
```diff
@@ -1,5 +1,6 @@
 ( function( $ ){
 
+	var ulrPrefix = $(".auto-slug").attr("data-ulrPrefix");
```
DominicWatson: I'd move this down to within the codeblock that starts on line 8. While not likely, there is a potential for multiple auto slug fields.

Also, we use these auto slugs for more than just pages in the tree, so we'd need a way to know to use this feature only when editing _pages_.

system/assets/js/admin/presidecore/formFields.js
```diff
@@ -7,14 +8,56 @@
...
+			URLslug(slug,ulrPrefix);
+		} );
+		$this.keyup( function(e){
+			var slug = $this.val().replace( /\W/g, "-" ).replace( /-+/g, "-" ).toLowerCase();
+			$this.val( slug );
+			URLslug(slug,ulrPrefix);
 		} );
 	});
 
+	function URLslug(slug,prefix){
```
DominicWatson: Similarly, this function needs to sit within the closure that starts on line 8 and refer to `$this`, rather than `$(".auto-slug")`. Otherwise, you cannot have multiple auto slug fields in a single page.

system/assets/js/admin/presidecore/formFields.js
```diff
@@ -7,14 +8,56 @@
...
+		$(".auto-slug").nextAll("span").first().text('');
+		
+		if(slug == 'msg'){
+			$(".auto-slug").after("<span style='color:red;'>"+prefix+"</span>");
+			$(".auto-slug").attr('disabled', true);
+		} else if (slug != ''){
+			$(".auto-slug").attr('disabled', false);
+			$(".auto-slug").after("<span><b>URL slug</b>: http://"+prefix + slug+".html</span>");
```
DominicWatson: The protocol should sit within the prefix (a site can be set to use https so it shouldn't be hard coded here)

system/assets/js/admin/presidecore/formFields.js
```diff
@@ -7,14 +8,56 @@
...
+		}
+	}
+
+	$("#parent_page").change(function(){
```
DominicWatson: This needs to sit within the anonymous function that starts on line 8 and change `$(".auto-slug")` to `$this`  (see comments above)

system/forms/preside-objects/page/edit.xml
```diff
@@ -11,7 +11,7 @@ This form is used as the base "edit page" form for Sitetree pages. See also [[fo
 		<fieldset id="main" sortorder="10">
 			<field sortorder="5" binding="page.parent_page" control="sitetreePagePicker" required="true" />
 			<field sortorder="10" binding="page.title" />
-			<field sortorder="20" binding="page.slug" required="true" />
+			<field sortorder="20" binding="page.slug" required="true" control="autoslug"/>
```
DominicWatson: This is deliberately _not_ an auto slug so that the page URL is not unintentionally changed with changes to the page title. I don't think we should change this behaviour.

DominicWatson:
This is a great start but there's a little bit more to it. The main pointers:
1. This should only happen when editing/adding a site tree page - the autoslug can be used in other forms and wouldn't want to have this happen
2. The edit page of a site tree page does not use "autoslug" to avoid URLs of pages being changed unintentionally - the js should work independently of autoslug. Could add `extraClasses="slug-preview"` to the form field definitions and bind the behaviour to that class instead
3. The js needs some tidying up to maintain modularity (see comments) :)
4. The base URL used needs more thorough calculating, I think something like this should do it:
   
   var parentPageUrl = event.buildLink( page=parentPageId );
   var baseUrl            = ReReplace( parentPageUrl, ".html$", "/" );

Dominic

cfmitrah:
@DominicWatson  Addressed your review comments, please check & confirm, if you need any changes on this.

DominicWatson:
Thanks for the updates @cfmitrah. I think there's still too much going on here though. We still need to ensure that this only applies to slug fields in site tree pages (they are used for other objects too). We could add an option to the slug field such as `<field ... siteTreePageSlug="true"/>` and use that to add whatever classes, etc. are necessary to tell the js to add this behaviour.

Also, the prefix for the slug should be made with the code:

```
var parentPageUrl = event.buildLink( page=parentPageId );
var baseUrl = ReReplace( parentPageUrl, ".html$", "/" );
```

This will give you the full prefix so no need to do anything with JavaScript to make that work (other than update the prefix when the parent page changes - can be an ajax call that makes use of the code above).

DominicWatson:
I'm going to close this one for now - it needs a different approach I think.

PR #75, opened 2015-12-04
presidecms-246 Added configurable class option to form controls

cfmitrah:


DominicWatson:
Lovely :) Manually Merging this into upcoming release-10.4.0

PR #76, opened 2015-12-04
PRESIDECMS-241 - Added the thumbnails on Asset manager List

cfmitrah:


DominicWatson:
Thanks @cfmitrah, this has been merged into upcoming 10.4.0 release. I tweaked it a little too to just show the thumbnail next to the title rather than a dedicated column in the table.

PR #77, opened 2015-12-09
PresideCMS-250 :: Asset Manager's Root folder label in the breadcrumb

cfmitrah:
Fixed the Asset Manager's Root folder label in the breadcrumb issue


DominicWatson:
Thanks @cfmitrah, merged into upcoming release :)

PR #79, opened 2015-12-10
PRESIDECMS-219 Link builder disable antispam option

nodoherty:
Initial development of option to allow users to bypass the mailto link encoding so links can be output as standard mailto: hrefs for content that is to be used in email content and other platforms that do not support or automatically cleanse Javascript.


DominicWatson:
All shades of perfect, merging into the upcoming 10.4.0 release. Thanks @nodoherty :)

PR #81, opened 2015-12-21
Fixed - PRESIDECMS-251 (Removed the extra white spaces)

josephpbuarao:
Hi @DominicWatson ,

Good day! I just added output=false to buildLink method on the RequestContentDecorator.cfc and then the line break issue is gone :)

Thanks and have a  great day ahead.


DominicWatson:
Nice one @josephpbuarao and good spot.

PR #86, opened 2016-01-04
URL Redirect Rule 500 server error issue

josephpbuarao:
Hi @DominicWatson ,

Just fixed the issue that causing the 500 server error when trying the add URL rewrite rule that already exist. Just added the following on the URLRedirects service.

property name="messageBox"          inject="coldbox:plugin:messageBox";

Before: http://screencast.com/t/Zj7Usj5O7
Now: http://screencast.com/t/XNoNwvTs

Thank you!


DominicWatson:
Thanks @josephpbuarao. This has been manually merged into upcoming 10.5.0 release.

PR #91, opened 2016-01-11
PRESIDECMS-264: Validation in REST framework

jjannek:
- extended REST Endpoint metadata evaluation to included information about required parameters and parameter types
- extended PresideRestService to do basic REST request parameter validation (evaluation of required parameters and parameter types)
- introduced 2 new events that can be intercepted: onMissingRequiredRestRequestParameter and onInvalidRestRequestParameterType


jjannek:
to be discussed:
- should we return the first validation error only or multiple errors at once?
- should we return different types of errors all at once, e.g. missing required params together with invalid param type validation errors

Right now it's implemented as follows:
First the list of required params are checked. an error message for each missing param is returned. this leads to an array of validation error messages
If there are param type validation errors as well, those are not returned in this case.
If all required params are set, then the type validation takes place. again multiple errors could be returned, one for each invalid parameter type.

DominicWatson:
This is great. I think a couple of things:
1. We should throw up all the errors we can discover (required fields, invalid types + any custom validation we may implement later).
2. The default implementation should come with its own opinion of what that error response should look like. Developers can then override that with custom interception points

Finally, we should look at using the already existing validation engine for this. I don't think it would be too much change to your code. Here is some rough pseudoish code to point in the right direction:

Note use of `ValidationEngine` below; this can be injected with Wirebox.

```
// for each verb:
var rulesetName = "restparamruleset.#restResourceName#.#verb#"; // or similar
var rules       = [];

for( param in functionMeta.parameters ) {
    if ( functionMeta.parameters[ param ].required ) {
        rules.append( { 
              fieldName = param
            , validator = "required" 
        } );
    }
    if ( _isTypeCheckable( functionMeta.parameters[ param ].type ) ) {
        // there actually isn't a 'isTypeOf' validator
        // implemented. We could add one, or we could
        // use diff validators depending on type. See `CoreValidators.cfc` 
        // for a place to add new ones
        rules.append( { 
              fieldName = param
            , validator = "isTypeOf"
            , params    = { type=functionMeta.parameters[ param ].type } 
        } );
    }
}
validationEngine.newRuleset( name=rulesetName, rules=rules );

// then store the ruleset name against the resource / verb combo
```

Then, when needing to validate, can do:

```
var validationResult = validationEngine.validate( rulesetName, passedData );
if ( !validationResult.validated() ) {
    // etc. validationResult object has methods like
    // validationResult.listErrorFields()
    // validationResult.getError( fieldName )
    // validationResult.fieldHasError( fieldName )
    // validationResult.getMessages()
}
```

Can chat on Slack to further this

DominicWatson:
Awesome stuff, thanks Jan. This has been manually merged into the upcoming 10.5.0 release branch :)

PR #94, opened 2016-01-18
PRESIDECMS-290 link.email_anti_spam might be empty string and this ch…

seakchiew:
…ecks if email_anti_spam is boolean and return false if it is not.


DominicWatson:
Nice one, thanks Seak Chiew. This has been merged into upcoming release :)

PR #95, opened 2016-01-20
PRESIDE-292 remove event.getBaseUrl() that causes duplicate site root…

seakchiew:
remove event.getBaseUrl() that causes duplicate site root URL


DominicWatson:
Nice one thanks @seakchiew - this has been manually merged into the upcoming 10.5.0 release

PR #97, opened 2016-01-27
PRESIDECMS-293 Get Error when change preside objects

cfmitrah:
While working on PRESIDECMS-293 ticket, We've found one issue while change datatype of the exists fieldand also fixed the issue.


DominicWatson:
Nice one - merging.

PR #100, opened 2016-01-29
Feature presidecms 273

cfmitrah:
We've modified a condition and the default value of maxValue , to fix this issue.


cfmitrah:
In this pull request some commits are showing. So i will create new pull request for PRESIDECMS-273 ticket issue and close this PR.

PR #104, opened 2016-01-29
PRESIDECMS-273

cfmitrah:
Default max value on spinner control is 999 - not a sensible default


DominicWatson:
Thanks Mani, merged into upcoming release :)

PR #105, opened 2016-02-01
PRESIDECMS-248 Active/Inactive/Date controlle state in sitetree

cfmitrah:
Green watch Icon and Red watch Icon shows in the page based on date range.


DominicWatson:
Thanks Mani, I've made a few tweaks but all now merged into upcoming release.

Changes seen here: https://github.com/pixl8/Preside-CMS/commit/961cc62854e162cedd1f73a0ba343425465dced8

PR #106, opened 2016-02-01
Feature presidecms 206

cfmitrah:
Removed the Timestamp function in CoreValidators.cfc and also made the necessary changes on PresideFieldRuleGenerator.cfc


DominicWatson:
Thanks Mani, merged into he upcoming release

PR #107, opened 2016-02-01
PRESIDECMS-249 Forward to Sitetree after saving a page

cfmitrah:
After saving a page, automatically go to the sitetree list.


DominicWatson:
I'm going to park this one for now, this was changed _from_ this behaviour to the current behaviour. Ideally there should be a "view only" screen for a page and this is where you should be taken.

PR #108, opened 2016-02-03
PRESIDECMS-295 

cfmitrah:
I've fixed the alert issue when navigating away from page.


DominicWatson:
I'm not 100% sure on this one. I think the issue that we've been facing is that the logic that stores the form state is running _before_ the richeditor fully loads and that richeditor adds more helper fields into the form. 

I think the real trick here is actually to store the field names as they were on original page load and ignore any added fields by javascript (as these will just be helper fields). 

Just waiting for ajaxStop I think might cause other issues.

PR #111, opened 2016-02-15
PRESIDECMS-297 Grid field option for Site tree page

cfmitrah:
Fixed the grid field option not working in site tree child page.


DominicWatson:
Awesome work, thank you. This has been merged into upcoming release.

PR #113, opened 2016-02-22
Feature presidecms 313

cfmitrah:
Done the ImageMagick configuration


system/services/assetManager/AssetManagerService.cfc
```diff
@@ -972,6 +972,13 @@ component displayName="AssetManager Service" {
 		var filename        = arguments.assetId & ( Len( Trim( arguments.versionId ) ) ? ".#arguments.versionId#" : "" ) & ".#fileext#";
 		var derivativeSlug  = ReReplace( arguments.derivativeName, "\W", "_", "all" ) & "_" & signature;
 		var storagePath     = "/derivatives/#derivativeSlug#/#filename#";
+		var assetDimension  = getAsset( id=arguments.assetId, throwOnMissing=true, selectFields=[ "height",'width' ] );
+
+		var assetProperty               = {};
+		assetProperty.height            = assetDimension.height ?: 0;
+		assetProperty.width             = assetDimension.width  ?: 0;
+		assetProperty.sourceFile        = expandPath( '/uploads/assets' & asset.storage_path );
```
DominicWatson, line 980: This is not correct. We cannot know that assets will live at `/uploads/assets` due to the possibility of multiple configured storage providers in the asset manager.

We need to do something like:

```
assetManagerService.getPhysicalPath( assetId );
```

This _new_ function would need to get the storage provider for the asset and call a `getPhysicalPath()` method on the asset _provider_ (this would also be a new method for the storage provider interface). For the file system storage provider, this would be a simple case of returning the file path combined with the storage path. For something like an S3 storage provider, we might need to implement something like a local cache directory for pulling down the file locally when we need access to a local file.

We should also pass `sourceFile` as a separate argument to any asset transformations. We should probably do this _instead_ of passing the asset binary.

system/services/assetManager/ImageMagickWrapper.cfc
```diff
@@ -0,0 +1,173 @@
+component output="false" displayname="ImageMagick"  {
```
DominicWatson, line 1: Could you remove all the `output=false` in the CFC? I realized too late that these are not needed in script based CFCs so there are lots of them in the codebase. But new code should not have `output=false` on script based components or their functions.

system/views/formcontrols/assetFolderPicker/index.cfm
```diff
@@ -1,7 +1,6 @@
 <cfscript>
 	inputName    = args.name         ?: "";
 	inputId      = args.id           ?: "";
-	inputClass   = args.class        ?: "";
```
DominicWatson, line 4: Why has this been removed?

system/views/formcontrols/assetPicker/index.cfm
```diff
@@ -1,7 +1,6 @@
 <cfscript>
 	inputName          = args.name             ?: "";
 	inputId            = args.id               ?: "";
-	inputClass         = args.class            ?: "";
```
DominicWatson, line 4: Why has this been removed?

system/services/assetManager/ImageMagickWrapper.cfc
```diff
@@ -0,0 +1,173 @@
...
+	public binary function resize(
+		  required binary  asset
+		,          numeric width               = 0
+		,          numeric height              = 0
+		,          string  quality             = "highPerformance"
+		,          boolean maintainAspectRatio = false
+		,          struct  assetProperty       =  {}
+	) output=false {
+		var image              = "";
+		var currentImageInfo   = {};
```
DominicWatson, line 21: Is this `assetProperty` argument really required? Is it there just to save the overhead of reading width / height from image file? 

If so, it should probably be called `assetProperties` and the `NativeImageManipulationService` should also use it.

cfmitrah:
Because of some merge conflicts, there are some undesired files also get committed. I will close this PR & create a new one with proper changes.

PR #115, opened 2016-02-24
Feature presidecms 258

cfmitrah:


system/handlers/admin/DataManager.cfc
```diff
@@ -344,6 +344,76 @@
 			);
 		</cfscript>
 	</cffunction>
+	<cffunction name="batchEditField" access="public" returntype="void" output="false">
+		<cfargument name="event"             type="any"     required="true" />
+		<cfargument name="rc"                type="struct"  required="true" />
+		<cfargument name="prc"               type="struct"  required="true" />
+		<cfscript>
+			var objectName   = rc.object;
```
DominicWatson: A number of issues with the code block below:
- Variables with no `var` scope (`fieldAttributes`, `fieldType`, `formControl`).
- `formControl` - let's look for a better solution here, ideally we should have something like: `renderFormControlForObjectField()`; I can help out here
- `manyToManySelect` special case: we really shouldn't need this. Need to understand what the problem is in order to find the right fix
- Breadcrumb title: need a title especially for "Batch edit". So: `uri="cms:datamanager.batchedit.breadcrumb.title", data=[ objectTitle, fieldTitle ]` (where object title and field title are translated labels for the object and field

system/i18n/cms.properties
```diff
@@ -83,6 +83,7 @@ datamanager.addrecord.title=Add {1}
 datamanager.addrecord.breadcrumb.title=Add record
 datamanager.addrecord.btn=Add {1}
 datamanager.cancel.btn=Cancel
+datamanager.bulkEdit.title=Bulk editing({3})  {1} > '{2}
```
DominicWatson: Missing a ' on the end?

system/i18n/cms_de.properties
```diff
@@ -70,6 +70,7 @@ datamanager.addrecord.title={1} hinzuf\u00FCgen
 datamanager.addrecord.breadcrumb.title=Datensatz hinzuf\u00FCgen
```
DominicWatson, line 70: Please do not provide German translations this way. We need to do this through Zanata and our translation processes (you do not need to worry about it)

system/i18n/cms_fr.properties
```diff
@@ -70,6 +70,7 @@ datamanager.addrecord.title=Ajouter des {1}
 datamanager.addrecord.breadcrumb.title=Ajouter un fichier
```
DominicWatson, line 70: Please do not provide french translations - we will do this through Zanata.

system/services/admin/DataManagerService.cfc
```diff
@@ -86,6 +86,17 @@ component output="false" singleton=true {
 		return ListToArray( fields );
 	}
 
+	public array function listBatchEditableFields( required string objectName ) output=false {
+		var objectAttributes = _getPresideObjectService().getObjectProperties(objectName);
+		var fields           = [];
+		for(property in objectAttributes){
+			if(StructKeyExists(objectAttributes[property],"batcheditable")){
```
DominicWatson: The logic here needs to be
- `batcheditable` attribute present on the property _and_ set to `true`
- field has no unique indexes
- field is not a `one-to-many` field

At the moment the code is just checking for the presence of the `batchEditble` attribute. Please add the extra logic.

system/services/admin/DataManagerService.cfc
```diff
@@ -227,6 +238,47 @@ component output="false" singleton=true {
...
+		  required string  objectName
+		, required string  sourceIds
+		, required string  updateValue
+		, required string  updateField
+		, required string  DataColumn
+		, required string  attributeRelationship
+		,          string  overwrite      = "append"
+	) output=false {
+
+		for( sourceID in sourceIds ) {
```
DominicWatson: `var` scope on `sourceId` (also please use `Id` rather than `ID`)

system/services/admin/DataManagerService.cfc
```diff
@@ -227,6 +238,47 @@ component output="false" singleton=true {
...
+			if( attributeRelationship == "many-to-many" ) {
+				multiSelectedValue = updateValue;
```
DominicWatson: `var` scoping issues throughout this function

system/services/admin/DataManagerService.cfc
```diff
@@ -227,6 +238,47 @@ component output="false" singleton=true {
...
+				if( overwrite != "overwrite") {
+					var previousData = _getPresideObjectService().getDeNormalizedManyToManyData( objectName = objectName, id = sourceID );
+					data[ sourceID ] = ListRemoveDuplicates( listAppend( multiSelectedValue, previousData[ DataColumn ] ) );
+				}else{
+					data[ sourceID ] =  multiSelectedValue;
+				}
+					result.multiSelect   = _getPresideObjectService().syncManyToManyData(  sourceObject   = objectName
```
DominicWatson: Indentation is too deep here. The pattern should look like this:

```
    functionCallResult = functionCall(
          arg1 = something
        , arg2 = somethingElse
    );
```

system/services/admin/DataManagerService.cfc
```diff
@@ -227,6 +238,47 @@ component output="false" singleton=true {
...
+			}else {
+				if ( attributeRelationship  == "many-to-one"){
+					singleSelectedValue     = updateValue;			
+					formData[ DataColumn ]  = singleSelectedValue;
+				} else {
+					singleSelectedValue     = updateValue;
+					formData[ updateField ] = singleSelectedValue;
+				}
+				
+				result.singleSelect = _getPresideObjectService().updateData( objectName  = objectName
```
DominicWatson: Indentation and alignment issues (as above).

system/handlers/admin/DataManager.cfc
```diff
@@ -344,6 +344,76 @@
...
+			event.setView( view="/admin/datamanager/batchEditField");
+		</cfscript>
+	</cffunction>
+
+	<cffunction name="batchEditAction" access="public" returntype="void" output="false">
+		<cfargument name="event"             type="any"     required="true" />
+		<cfargument name="rc"                type="struct"  required="true" />
+		<cfargument name="prc"               type="struct"  required="true" />
+		<cfscript>
+			rc.updateValue           = rc[rc.updateField];
```
DominicWatson: Feedback on code block below:
- `var` scope local variables
- Do not use `rc` for variables that are calculated by code. `rc` should be reserved for user input only. Use either local `var` scope, for variables that are only needed within the function, or `prc` for variables that need to be accessed later in views, etc.
- Do not assume variables exist, e.g. `rc.updateField`. Instead:

```
var updateField = rc.updateField ?: "";
// etc.
```

system/services/admin/DataManagerService.cfc
```diff
@@ -227,6 +238,47 @@ component output="false" singleton=true {
 
 		return result;
 	}
+	public struct function batchSaveFieldChanges(
+		  required string  objectName
+		, required string  sourceIds
+		, required string  updateValue
+		, required string  updateField
+		, required string  DataColumn
```
DominicWatson: should be "headless" camel case, i.e. `dataColumn` _not_ `DataColumn`. Also, intention of this variable is not clear in the name; `forekgnKeyField` may be clearer

system/services/admin/DataManagerService.cfc
```diff
@@ -227,6 +238,47 @@ component output="false" singleton=true {
...
+					formData[ DataColumn ]  = singleSelectedValue;
+				} else {
+					singleSelectedValue     = updateValue;
+					formData[ updateField ] = singleSelectedValue;
+				}
+				
+				result.singleSelect = _getPresideObjectService().updateData( objectName  = objectName 
+																            , data       = formData 
+																            , id         = sourceId  );
+				structClear(formData);
```
DominicWatson: Rather than this, it would be cleaner to do:

```
for( var sourceId in sourceIds ) {
    var formData = {};

    // ...
}
```

This also fixes the fact that `formData` is not `var` scoped.

system/services/admin/DataManagerService.cfc
```diff
@@ -227,6 +238,47 @@ component output="false" singleton=true {
...
+			if( attributeRelationship == "many-to-many" ) {
+				multiSelectedValue = updateValue;
+				if( overwrite != "overwrite") {
+					var previousData = _getPresideObjectService().getDeNormalizedManyToManyData( objectName = objectName, id = sourceID );
+					data[ sourceID ] = ListRemoveDuplicates( listAppend( multiSelectedValue, previousData[ DataColumn ] ) );
+				}else{
+					data[ sourceID ] =  multiSelectedValue;
+				}
+					result.multiSelect   = _getPresideObjectService().syncManyToManyData(  sourceObject   = objectName
+											                                 , sourceProperty = DataColumn
```
DominicWatson: I'm fairly certain that `sourceProperty` here should be `updateField`.

system/views/admin/datamanager/batchEditField.cfm
```diff
@@ -0,0 +1,35 @@
...
+	recordLabel         = rc.pickField    ?: "";
+	objectTitleSingular = translateResource( uri="preside-objects.#object#:title.singular", defaultValue=object ?: "" );
+	editRecordTitle     = translateResource( uri="cms:datamanager.bulkEdit.title", data=[ LCase( objectTitleSingular ),recordLabel,ListLen(rc.id) ] );
+	saveButton          = translateResource( uri="cms:datamanager.savechanges.btn", data=[ LCase( objectTitleSingular ) ] );
+	prc.pageIcon        = "pencil";
+	prc.pageTitle       = editRecordTitle;
+</cfscript>
+<cfoutput>
+	<form class="form-horizontal" method="post" data-dirty-form="protect" action="#event.buildAdminLink( linkTo='datamanager.batchEditAction')#">
+			<input type="hidden" name="sourceIds"   value="#rc.id#">
```
DominicWatson: Watch the indentation here, there is a double indent here where there should only be one level.

DominicWatson:
Nice one, this has been merged manually and tweaked. Will appear in upcoming 10.5.0 release.

PR #117, opened 2016-02-25
Feature presidecms 265 formbuilder

cfmitrah:


system/services/formbuilder/FormBuilderValidationService.cfc
```diff
@@ -124,6 +124,16 @@ component {
 	public any function validateFormSubmission( required array formItems, required struct submissionData ) {
 		var ruleset = getRulesetForFormItems( items=arguments.formItems );
 
+		for ( item in arguments.formItems ) {
+			if ( isdefined( "item.configuration.name" ) && isdefined( "item.type.id" ) ) {
+				var field = item.configuration.name;
+				if ( item.type.id == 'matrix' && arguments.submissionData["#field#"] != "" ) {
```
DominicWatson, line 130: We need to not have specific workarounds within the core code. The approach we take must be generic to allow new item types to take similar approaches without needing to change the core code.

PR #118, opened 2016-03-02
PRESIDECMS-293 Get Error when change preside objects

cfmitrah:
Fixed the issue while changing properties in preside-object.


DominicWatson:
Brilliant, thanks for this Mani :)

PR #120, opened 2016-03-09
set no action for on update and on delete

locrasia:
For ms sql I think just use only no action for onupdate and ondelete


DominicWatson:
This is a bad idea - cascading deletes are a design choice and the system will start breaking in unexpected ways if we just ignore them. We should stick with the approach created that only sets no-action when we need it to (i.e. if we find that SQL server complains about recursive cascades, then set `ondelete="cascade-if-no-cycle-check"`.

Closing this PR for now.

PR #121, opened 2016-03-09
Feature presidecms 199

cfmitrah:


DominicWatson:
Nice one, thanks for this :) Could you:
- Remove the French and German translations (we deal with this over at https://translate.zanata.org/zanata/iteration/view/presidecms/10.5.0)
- Provide a couple of screenshots (I'm stretched at the moment and that would really help me review)

TIA

cfmitrah:
I've removed the French and German translations and the screenshots are below
alloptionselected
selectall options

DominicWatson:
Looking good :). I wonder could we have a slightly different visual treatment on the "Select all" and "Select none" entries? Perhaps italic? Or does a font awesome icon render in a dropdown? e.g. (`<i class="fa fa-fw fa-check"></i>`).

Also, the german translation is still there ;)

cfmitrah:
I've added text color for "Select all" and "Select none" to differentiate. And also I've removed german translation. Please refer the screen shot.
selectall

DominicWatson:
This is getting there, I'm still not entirely convinced by this one though so going to leave for 10.7.0. My concern lies with selects that could be selecting against 1000s of records which are mostly searched over ajax. The current implementation won't select all of those and, if it did, the page would likely become very slow indeed.

Some other ideas might be to have some special marker to say that "all" records are selected. Or, another option is to have this as a feature flag for the uber select that is turned off by default. Then, sites are free to use it on object pickers where it makes sense to do so.

DominicWatson:
I'm going to be harsh and close this along the lines of "If in doubt, leave it out.". Maybe we can revisit should this be a priority for someone.

PR #122, opened 2016-03-10
PRESIDECMS-339: fixed main navigation to have dropdowns working

jjannek:


system/views/core/navigation/mainNavigation.cfm
```diff
@@ -5,18 +5,18 @@
...
+	aHasChildrenClass       = args.aHasChildrenClass       ?: 'dropdown-toggle';
+	aHasChildrenAttributes  = args.aHasChildrenAttributes  ?: 'data-toggle="dropdown"';
 </cfscript>
 
 <cfoutput>
 	<cfloop array="#menuItems#" index="i" item="item">
-		<cfset hasChildren = item.children.len() />
-		<li class="<cfif item.active>#liCurrentClass#</cfif><cfif hasChildren> #liHasChildrenClass#</cfif>" <cfif hasChildren>#liHasChildrenAttributes#</cfif>>
-			<a class="<cfif item.active>#aCurrentClass#</cfif><cfif hasChildren> #aHasChildrenClass#</cfif>" href="#event.buildLink( page=item.id )#" <cfif hasChildren>#aHasChildrenAttributes#</cfif>>#item.title#</a>
+		<cfset hasChildren = not item.children.isEmpty() />
```
DominicWatson: This change makes me curious, what does `not item.children.isEmpty()` give us that `item.children.len()` does not? (other than a "true" boolean rather than an integer)?

jjannek: Sorry about that one, it slipped through.

I tried to understand what went wrong regarding not displayed markup and completely changed the file. Afterwards I reverted most of what I did so that i ended up with the least possible changes - and forgot to change that line back.

From a coding style and with a Java / strongly typed language background, I am not used to the concept of treating non-zero numbers the boolean way as true. It's a strange concept imho when you "read" the code, e.g. i always used "if (myarray.len() gt 0) {...}" instead of "if (myarray.len()) {...}" because it reads more like natural language. But that's only a personal opinion and I try to accommodate myself with this style of coding as most of the CF-developers do it anyway. :-)

The explanation was just to satisfy your curiosity. I will revert the line to what it was before.

DominicWatson:
Awesome, thanks @jjannek  - this has been manually merged into upcoming 10.6.0 release.

PR #123, opened 2016-03-15
Feature presidecms 82

cfmitrah:


system/assets/css/admin/specific/auditTrail/index/auditTrail.less
```diff
@@ -0,0 +1,126 @@
+.message-item {
```
DominicWatson, line 1: One little thing to consider is to use more nesting in LESS files. Really helps organise the styling. See http://lesscss.org/

system/assets/js/admin/specific/auditTrail/index/auditPagination.js
```diff
@@ -0,0 +1,56 @@
+( function( $ ){
```
DominicWatson: We can move this to `/js/admin/presidecore` (name it preside.load.more.js). Its pretty small and I'm sure we will use it again.

system/config/Config.cfc
```diff
@@ -222,6 +224,7 @@ component output=false {
 		settings.adminRoles.contentadmin       = [ "sites.*", "presideobject.site.*", "presideobject.link.*", "sitetree.*", "presideobject.page.*", "datamanager.*", "assetmanager.*", "presideobject.asset.*", "presideobject.asset_folder.*", "formbuilder.*", "!formbuilder.lockForm", "!formbuilder.activateForm" ];
 		settings.adminRoles.contenteditor      = [ "presideobject.link.*", "sites.navigate", "sitetree.*", "presideobject.page.*", "datamanager.*", "assetmanager.*", "presideobject.asset.*", "presideobject.asset_folder.*", "!*.delete", "!*.manageContextPerms", "!assetmanager.folders.add" ];
 		settings.adminRoles.formbuildermanager = [ "formbuilder.*" ];
+		settings.adminRoles.auditTrailmanager  = [ "auditTrail.*" ];
```
DominicWatson: Its good that you've made this work. However, I don't think that we need / want a specific role just for the audit trail. Instead, add "auditTrail.*" to the "sysadmin" admin role.

system/handlers/admin/AuditTrail.cfc
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	public void function index( event, rc, prc ) output=false {
+		prc.logs         = getModel( "AuditService" ).getAuditLog(1,10);
+		prc.pageTitle    = translateResource( "cms:auditTrail.page.title" );
+		prc.pageSubTitle = translateResource( "cms:auditTrail.page.subtitle" );
+	}
+
+	public any function loadMore(event, rc, prc) output=false {
+		prc.logs         = getModel( "AuditService" ).getAuditLog(rc.start,10);
```
DominicWatson: We need to be more _defensive_ with the `rc.start` here. Whenever there is raw user input, we need to ensure it is as expected. Also, please use "page", rather than start. So:

```
var page = Val( rc.page ?: 2 );
prc.logs = getModel( "AuditService" ).getAuditLog( page, 10 );
```

system/handlers/admin/AuditTrail.cfc
```diff
@@ -0,0 +1,44 @@
...
+	private void function _permissionsCheck( required string key, required any event ) {
+		var permKey   = "auditTrail." & arguments.key;
+		var permitted = hasCmsPermission( permissionKey=permKey );
+
+		if ( !permitted ) {
+			event.adminAccessDenied();
+		}
+	}
+
+	public void function viewAuditTrail( event, rc, prc ) {
```
DominicWatson: This should be named `viewLog` and moved _above_ the "PRIVATE UTILITY" section.

system/i18n/cms_de.properties
```diff
@@ -768,3 +768,8 @@ urlRedirects.ruleNotFound.error=Die Regel wurde nicht gefunden
 urlRedirects.editRule.pageTitle=Regel bearbeiten
 urlRedirects.editRule.pageSubtitle={1}
 urlRedirects.editRule.breadcrumbTitle=Regel bearbeiten
+
```
DominicWatson: Please leave translations to our translators to save yourself effort. This needs to go through our translation workflow in Zanata.

system/services/admin/AuditService.cfc
```diff
@@ -1,11 +1,21 @@
...
 
 // CONSTRUCTOR
 	/**
 	 * @dao.inject presidecms:object:audit_log
+	 * @coldboxController.inject  coldbox
 	 */
-	public any function init( required any dao ) output=false {
+	public any function init(
+		  required any dao
+		, required any coldboxController
```
DominicWatson: Because this service is tagged as `@presideService`, we do not need to inject `coldboxController` here. Instead, we can just use `$getColdbox()`. See https://docs.presidecms.com/devguides/presidesuperclass.html

system/services/admin/AuditService.cfc
```diff
@@ -32,6 +42,49 @@ component output="false" singleton=true {
 		} );
 	}
 
+	public query function getAuditLog(
+		  numeric startRow = 1
+		, numeric maxRows  = 0
+		) output=false {
```
DominicWatson: The closing `)` should align with the beginning of the function definition. Also, `output=false` is not required for script based components (even though lots of old code has it - we need to start removing them when we find them!). So:

```
public query function getAuditLog(
      numeric startRow = 1
    , numeric maxRows  = 0
) {
```

system/services/admin/AuditService.cfc
```diff
@@ -32,6 +42,49 @@ component output="false" singleton=true {
...
+		var records = _getDao().selectData(
```
DominicWatson, line 44: Indentation is a little off here. Looks like two tabs when only one required. Also, closing `)` should align with the `var`, so:

```
var records = _getDao().selectData(
      selectFields = [ "audit_log.id", "audit_log.type", "audit_log.datecreated", "audit_log.action", "security_user.email_address", "security_user.known_as" ]
    , orderby      = "audit_log.datecreated desc"
    , startRow     = arguments.startRow
    , maxRows      = arguments.maxRows
);
```

system/services/admin/AuditService.cfc
```diff
@@ -32,6 +42,49 @@ component output="false" singleton=true {
...
+	}
+
+	/**
+	 * Returns the auditTrail record matching the given ID
+	 *
+	 * @autodoc
+	 * @auditTrailId.hint The ID of the auditTrail you wish to get
+	 *
+	 */
+	public query function getAuditTrail( required string auditTrailId ) {
```
DominicWatson: I think a better name for this is `getAuditLog()` - "trail" suggests that you could get multiple logs. (also rename `auditTrailId` to `auditLogId`.

system/helpers/presideProxies.cfm
```diff
@@ -64,7 +64,9 @@
 		<cfreturn getSingleton( "notificationService" ).renderNotification( argumentCollection = arguments ) />
 	</cffunction>
 
-
+	<cffunction name="renderLogDetails" access="public" returntype="any" output="false">
```
DominicWatson: Can we rename this `renderAuditLog()`?

system/handlers/renderers/content/auditLogEntry.cfc
```diff
@@ -0,0 +1,33 @@
...
+	private string function default( event, rc, prc, args={} ) {
+		var auditTrailId  = args.id ?: ""
+		var records       = AuditService.getAuditTrail( auditTrailId );
+
+		return renderView( view="/renderers/content/auditLogEntry/default", args=args );
+	}
+
+	private string function login_success( event, rc, prc, args={} ) {
+		var auditTrailId  = args.id ?: ""
+		var records       = AuditService.getAuditTrail( auditTrailId );
```
DominicWatson: rather than `records`, this should be named `auditLog` (we expect a single audit log record here)

system/views/renderers/content/auditLogEntry/default.cfm
```diff
@@ -0,0 +1,21 @@
...
+		<tr>
+			<th>#translateResource( "cms:auditTrail.type" )#</th>
+			<td>#args.type#</td>
+		</tr>
+		<tr>
+			<th>#translateResource( "cms:auditTrail.detail" )#</th>
+			<td>#args.detail#</td>
+		</tr>
+		<tr>
+			<th>#translateResource( "cms:auditTrail.Date&Time" )#</th>
```
DominicWatson: Please make the i18n property name `auditTrail.dateAndTime` (do not use non alpha numeric characters in the names, other than `.`).

DominicWatson:
Hey Mani, overall this is looking really great, thank you. Please see feedback on code formatting and some variable / function naming issues, etc.

Finally, I think the renderer for each log type needs a little more thinking about. We should make a change to the data model and logging API so that developers can attach arbitrary data to a log, i.e. change detail to a `longtext` field in the database and enable developers to pass any data type. This would then be serialized to json before saving in the DB:

```
event.audit(
      detail = { objectName=objectName, recordId=recordId }
    , source = "datamanager"
    , action = "edit_record"
    , ...
);
```

The renderer for each type should then enable developers to supply custom _descriptions_ of the log event. So, where you have the data table for individual log view, the custom renderer would only be needed for the "detail" section. It should also be used to render the description when viewing the trail.

Let me know if you want to go through this together.

system/services/admin/AuditService.cfc
```diff
@@ -32,6 +35,47 @@ component output="false" singleton=true {
...
+		);
+	}
+
+	public any function renderLogMessage(  required struct log ) {
+
+		var viewletEvent = "renderers.auditLog.auditLogEntry." & arguments.log.action;
+		if ( $getColdbox().viewletExists( viewletEvent ) ) {
+			return $getColdbox().renderViewlet(
+				  event = viewletEvent
+				, args  = arguments.log);
```
DominicWatson: closing parenthesis belongs on a new line and should line with the beginning of the first line (return). So:

```
            return $getColdbox().renderViewlet(
                  event = viewletEvent
                , args  = arguments.log
            );
```

Same for function call below.

DominicWatson:
This has been merged into 10.7.0 (with some further work). Thanks guys,

PR #125, opened 2016-03-31
Fixed the Issue while copy the text using "Ctrl+C".

cfmitrah:


DominicWatson:
Awesome, thanks for this - merging into 10.6.0 :)

DominicWatson:
Quick note: I actually changed the approach here to use jQuery's event.ctrlKey, event.altKey, etc. to detect for modifiers in general. This is more robust, simpler and covers more bases.

See "Event properties" section here: http://api.jquery.com/category/events/event-object/

PR #126, opened 2016-04-05
Feature presidecms 363

cfmitrah:


DominicWatson:
Thanks for this, great effort. However, it's not quite the right approach. _All_ of our i18n translations should be in `.properties` files and _not_ in js. To allow translations to be available to javascript, put the keys in `cms.properties` and then, in js, do:

```
i18n.translateResource( 'cms:day.monday' );
```

etc. This will take some effort to do correctly for the date picker.

DominicWatson:
Hey, this is looking good, nearly there I think. However, we should take locale handling _out_ of the javascript altogether. Have the javascript just have a default locale that is populated using `translateResource()` calls in CMS. This way, language changes in the CMS will be reflected in the Javascript without needing to have hardcoded locale lists, etc. (i.e. fr, de).

cfmitrah:
@DominicWatson Not sure, where we have hardcoded locale list in JS, Could you please specific on line number & file name?

DominicWatson:
bootstrap.datepicker.js, Line ~840 and below sets up french and german.

DominicWatson:
I think this needs more careful looking at. I see some conflicts of approach and don't quite have the time to sort at the moment. Closing for now, but we should revisit.

Thanks for work on this.

PR #127, opened 2016-04-05
Relative Date in Formbuilder updated by offset field value

cfmitrah:


system/handlers/formbuilder/item-types/Date.cfc
```diff
@@ -13,13 +13,15 @@ component {
 
 			switch( args.relativeOperator ) {
 				case "lt":
-					args.maxDate = DateAdd( 'd', -1, Now() );
+					var diff = val(args.offset) ? val(-args.offset) : -1;
```
DominicWatson, line 16: I'm not sure of the default `-1` here (same for the "gt" offset). I would have thought:

```
if ( Val( args.offset ) ) {
    args.maxDate = DateAdd( 'd', Val( -args.offset ), Now() );
}
```

Can you check this?

system/i18n/formbuilder/item-types/date.properties
```diff
@@ -25,4 +25,7 @@ fieldset.relative.title=Relative dates
 relativeOperator.lt=Less than...
 relativeOperator.lte=Less than or equal to...
 relativeOperator.gt=Greater than...
-relativeOperator.gte=Greater than or equal to...
\ No newline at end of file
+relativeOperator.gte=Greater than or equal to...
+
+field.offset.title=Offset(Days)
```
DominicWatson: Should have space between `Offset` and `(Days)`. So `Offset (days)`. Help should be:

> For relative dates, supply an optional offset to allow date restrictions such as 14 days greater than today.

DominicWatson:
Thanks for this :) Please check review comments above.

system/handlers/formbuilder/item-types/Date.cfc
cfmitrah, line 16: They already set default value -1 in 'lt' (same for the 'gt'):
`args.maxDate = DateAdd( 'd', -1, Now() );`
I've replace this line with
`var diff = val(args.offset) ? val(-args.offset) : -1;`

DominicWatson, line 16: Ah yes, I'm being slow. Thanks for the clarification :)

DominicWatson:
Frantastic, thanks for this. This has been manually merged into 10.6.0

PR #128, opened 2016-04-05
PRESIDECMS-369 : Layout is not changing issue

cfmitrah:
Fixed the layout not changing issue for ems application.


DominicWatson:
Nice one, thanks for this. Pulling in to 10.6.0. In the future though, please just delete code rather than commenting it out :)

PR #129, opened 2016-04-06
Fixed Notification Center  shows badge after notification is deleted …

cfmitrah:


DominicWatson:
Thanks for this Selva. This fixes the symptom but changes the desired behaviour of dismissing a notification. The problem is actually a cache invalidation problem - reloading caches with `reload caches` from the dev console makes the notification badge go away.

What we _could_ try here is to manually delete _all_ of the consumer records first before deleting the parent notification record:

```
public numeric function dismiss( required array notificationIds ) autodoc=true {
    _getConsumerDao().deleteData( filter = { admin_notification=arguments.notificationIds } );

    return _getNotificationDao().deleteData( filter = { id=arguments.notificationIds } );
}
```

Could you test this approach and commit changes if working?

DominicWatson:
Nice one, thanks Selva. Manually merged into upcoming 10.6.0 :)

PR #132, opened 2016-04-12
Displayed system information(versions).

cfmitrah:
version
version1


DominicWatson:
Nice work Selva, thanks - merged into upcoming 10.6.0 release.

PR #133, opened 2016-04-13
visually indicated while working on non-live environment

cfmitrah:


DominicWatson:
Thanks for this Selva. I think an improvement might be simply to have a `settings.environmentMessage` setting that, if not empty, can be shown as you have done. So, then in config, you could do:

```
function configure() {
    settings.environmentMessage = ""
}

function local() {
    settings.environmentMessage = "You are currently viewing the application in a local environment";
}

function dev() {
    settings.environmentMessage = "You are currently viewing the application in the DEV environment";
}


function stagomg() {
    settings.environmentMessage = "You are currently viewing the application in the STAGING environment";
}
```

This would provide greater flexibility on the messaging used in various environments for any given application (see also http://coldbox.ortusbooks.com/content/configuration/coldboxcfc/configuration_directives/environments.html)

cfmitrah:
I have done **settings.environmentMessage** configuration for each environment as you mentioned above. Please check it and let me know.

DominicWatson:
Nice one, thanks Selva. This will be released with 10.6.0.

PR #137, opened 2016-04-19
The Lucee dot notation settings will update if not "Keep original case"

cfmitrah:


DominicWatson:
I'm still unsure on this one though I like the general approach of making things easier. A couple of things:
1. Abstract the logic into its own private function in `/preside/system/Bootstrap.cfc`, named `_ensureCaseSensitiveStructSettingsAreActive()` - call that function from the `_initEverything()` function (before the call to `_fetchInjectedSettings()`
2. Do a simple check to see if updating the setting is necessary. Something like:

```
var check = { sensitivity=true };
var caseSensitive = check.keyArray().find( "sensitivity" );

if ( !caseSensitive ) {
    // update the admin settings
}
```

DominicWatson:
I've added some safety checking in there and avoided calling the admin API unless necessary. Great work guys thank you. Merged into 10.7.0

PR #140, opened 2016-04-20
Added translation status column in preside object list.

cfmitrah:


system/views/admin/datamanager/_listingTranslations.cfm
```diff
@@ -0,0 +1,28 @@
+<cfparam name="args.translations"     type="array"  />
+<cfparam name="args.translateUrlBase" type="string" />
+
+<cfoutput>
+	<cfif !arrayIsEmpty( args.translations )>
+		<cfloop index="arrayIndex" from="1" to="3">
```
DominicWatson, line 6: I would expect this to break if there were only two languages (array index out of bounds). Can you verify this?

DominicWatson:
Hey Selva, this is looking fantastic. Thank you. I've just spotted a possible issue though. Could you check that having less than three configured languages does not break it?

DominicWatson:
Nevermind actually (apologies if you've started). I decided to make a few tweaks and fixed the issue in the process. Great work on this, thank you. Merging into upcoming release.

PR #141, opened 2016-04-21
Displayed loading sheen while switching the asset folders.

cfmitrah:


DominicWatson:
Great, thanks for this. Manually merging into upcoming release :)

PR #144, opened 2016-04-26
Checkboxes checked correctly.

cfmitrah:


DominicWatson:
Nice one, thanks for this. I have manually merged into upcoming release + made a few tidy ups.

PR #145, opened 2016-04-26
Added default sort button and @datamanagerSortOrders

cfmitrah:


system/handlers/admin/DataManager.cfc
```diff
@@ -46,16 +46,22 @@
 
 		<cfscript>
 			var objectName   = event.getValue( name="id", default="" );
+			if(structKeyExists(rc, "defaultSort") && rc.defaultsort ){
+				var cookieName = 'SpryMedia_DataTables_object-listing-table-'&objectName&'_';
+				cookie[cookieName] = { value="",expires="now",domain="127.0.0.1",preserveCase="yes"};
```
DominicWatson: I think this cookie behaviour is not right (particularly the hardcoded 127.0.0.1 domain ;)). Could we just remove these two lines? (IF we don't want the datatables to remember sortorders, we can tell it not to do that in javascript, but I think that's a separate issue).

DominicWatson:
Thanks for this effort guys. Going to close though as a duplicate.

PR #148, opened 2016-05-05
PRESIDECMS-397 Require full login field in adding/editing asset is dsplaying as textfield instead of toggle

ghost:
Require full login field not displaying correctly, suspect is cause by field not inheriting correctly when elasticsearch add searchenabled to **full_login_required** field. changing it to text field


DominicWatson:
Hi @nelsoncwz, thanks for the PR. I think this is an issue with either the elasticsearch extension or the application that is flagging this field as searchable and so should be fixed there. 

One thing for sure though is that this field should never be _searchable_ in the ElasticSearch extension meaning of that (i.e. it should _not_ be indexed with full text analyzers).

Am closing this and the related issue. Can you confirm that applying this fix to the application itself solves the problem?

PR #149, opened 2016-05-26
CMS-404 Now the Translation form should not only be accessible via th…

cfmitrah:
…e edit form of an object


DominicWatson:
Nice one thanks guys. WIll be released with 10.7.0

PR #150, opened 2016-06-01
PRESIDECMS-418 Allow for differing EXIF image rotations in asset oper…

patmoody:
…ations

First pull request


patmoody:
The two portrait types I have catered for on the Native image service are portrait_6 and portrait_8 on this project.
https://github.com/recurser/exif-orientation-examples/
I need some help with the dropzone changes at this point. I've made the javascript changes but have not been able to get them to work successfully.

system/services/assetManager/ImageMagickService.cfc
```diff
@@ -22,10 +22,19 @@ component displayname="ImageMagick"  {
 		,          string  quality             = "highPerformance"
 		,          boolean maintainAspectRatio = false
 	) {
-		var currentImageInfo = getImageInformation( arguments.asset );
+
+		var imageBinary = arguments.asset;
+		//var checkImageEXIF = $getPresideSetting( "asset-manager", "chk_imageEXIF" );
```
DominicWatson: Is there a good reason to _not_ check the exif? (i.e. do we need this to be configurable?)

system/services/assetManager/ImageMagickService.cfc
```diff
@@ -184,6 +208,37 @@ component displayname="ImageMagick"  {
 		throw( type="AssetTransformer.shrinkToFit.notAnImage" );
 	}
 
+	public binary function getImageRotation( required binary asset ) {
```
DominicWatson: I think this should be renamed to "autoCorrectImageOrientation()"

DominicWatson:
Hey @patmoody this is great :). I imagine that the issue with the javascript is that when the "plugins" directory gets bundled up and minified, the exif.js is coming _after_ dropzone.js. 

To fix this, find the relevant lines in /system/assets/Gruntfile.js (line 81-86) and add exif.js  before *.js to ensure it is included before dropzone.js. It would also be worth bumping the plugins version to plugins-1.4.000.min.js. After a `grunt all` you'd then need to manually remove plugins-1.3.000.min.js.

Let me know if that doesn't make sense.

patmoody:
Hi @DominicWatson.
I've made the suggested changes to the services and added an additional condition for portrait image type 3 - rotated 180 deg. 3, 6 & 8 seem to be the most likely.
https://github.com/recurser/exif-orientation-examples/
The check exif setting was whilst I was concerned about the processing time, particularly with the native service. Because the checks are completed before rotating, this should not be too great a performance hit (unless all of the users portrait images fall into these categories).
I've had one go at the js but will have to try again once I get grunt sorted out properly.

DominicWatson:
Awesome thanks @patmoody. This has been merged into 10.7.0 bleeding edge.

PR #155, opened 2016-06-16
PRESIDECMS-434

john5onCheng:
Add custom select fields option and set to get labelField only


DominicWatson:
Closing this as their is some similar work and thinking around this area that should make creating custom renderers for objects more easy.

Thanks for the PR though :)

PR #158, opened 2016-07-15
PRESIDECMS-454 updated navbar CSS updated.

cfmitrah:


DominicWatson:
Right now, we need to compile asset changes before committing. Are you able to do that?

DominicWatson:
Merged into upcoming release. Thanks selva :)

PR #160, opened 2016-07-27
PRESIDECMS-453: Fixed item selection issue on assetManager and dataMa…

cfmitrah:
…nager listing tables.


cfmitrah:
Color has been removed while deselecting the field
item_fix
asset_item_fix

PR #162, opened 2016-07-29
PRESIDECMS-487: Fixed margin issue in dateTime picker

cfmitrah:


cfmitrah:
Fixed the margin issue occurs in dateTime picker.
datetime_picker

PR #165, opened 2016-08-11
PRESIDECMS-490 : Removed "Choose files" button in asset upload window.

cfmitrah:


DominicWatson:
I'm going to close this as I disagree with the idea. I don't have a problem with multiple buttons in a page to achieve the same thing. 

I think the UI could be improved, but we need a more total approach to that I think.

PR #167, opened 2016-08-19
PRESIDECMS-82

john5onCheng:
Add the missing id field for permanentlyDeleteAsset function.
This is Branch out from 10.7


DominicWatson:
Unfortunately, this has been done in the wrong branch. It should have been done directly in the PRESIDECMS-82 branch that already exists and merged back to that branch. 

I'll manually make the changes myself but thanks for picking this up @john5onCheng.

PR #172, opened 2016-09-06
PRESIDECMS-534 : Include non clickable datatable rows

cfmitrah:


DominicWatson:
Nice one, thanks guys :)

PR #174, opened 2016-09-14
Feature presidecms 546 asset dropdown overlapping

nelsonchuah:
Prevent Attachment dropdown picker text overlapping each other when file's title excess 1 line


DominicWatson:
Thanks for this @nelsonpixl8 - I've manually applied your patch as their were issues with the PR (its including changes to /ckeditor folder for instance.

Closing, but accepted :).

PR #176, opened 2016-09-27
PRESIDECMS-563 Fix boolean queries to work with PostgreSQL

sebduggan:
Boolean values in SQL queries need to be surrounded by single quotes for cross-db compatibility


DominicWatson:
Great, thanks Seb. Having one of those hectic days so can't merge right now. There are a couple of PRs that can be made together so I'll create a new hotfix for both asap.

PR #177, opened 2016-09-28
PRESIDECMS-202: Fixed datePicker issue on time span filter

cfmitrah:


DominicWatson:
Great spot Selva, thank you!

PR #180, opened 2016-09-30
PRESIDECMS-461 : Added allowed operations for status column.

cfmitrah:


DominicWatson:
Thanks @cfmitrah. This isn't quite right though I think. More suitable would be to add "translate" to the allowedOperations set. Translate is an operation, it doesn't matter that it visually appears somewhere else.

closing

PR #181, opened 2016-10-03
PRESIDECMS-91 : Enable the SCAYT by default on richeditor.

cfmitrah:


DominicWatson:
Thanks for this @cfmitrah. We're going to close this though as we don't agree that this should be the default behaviour. An application can override if they like though.

PR #182, opened 2016-10-03
PRESIDECMS-141: display a PDF icon for password protected pdf preview.

cfmitrah:


DominicWatson:
Nice one @cfmitrah thanks for this. Merging into upcoming release

PR #183, opened 2016-10-04
PRESIDECMS-13: Admin can configure the count of login attempts.

cfmitrah:


DominicWatson:
Hey, somethings gone wrong here. Looks like how you've branched has meant that we're getting all the changes through from an entire release.

Always have a single feature branch per pull request, and always branch from stable.

Apologies but need to close this one (thanks for the effort of the feature though) - are you able to redo this in a new branch from stable?

PR #185, opened 2016-10-04
PRESIDECMS-300: Added distinct argument on all sql adapters for selec…

cfmitrah:
…t query.


DominicWatson (changes requested):
Hey guys, this looks good. However, it looks like its missing a way to specify a DISTINCT query in the presideObjectService selectData() method.

i.e.
```
presideObjectService.selectData(
      selectFields = [ ... ]
    , distinct     = true
    ...
);
```

Would be good to add this and test it working. Unless you think this works without that?

cfmitrah:
Added distinct argument in select data method and resolved conflicts.

DominicWatson:
Great, thanks @cfmitrah. Merging into 10.8.0 :)

PR #191, opened 2016-10-06
PRESIDECMS-140: Fixed image re size bug.

cfmitrah:


DominicWatson:
Nice one, thanks for this. Merged into upcoming 10.8.0 release

PR #192, opened 2016-10-06
PRESIDECMS-46: Added robots.txt conf under system config.

cfmitrah:


DominicWatson:
I'm not to sure on this one. I think the file storage, etc. is overkill. I'd rather we served this dynamically (clusters, etc. prove problematic for this kind of thing).

A simple system setting + custom route should do the trick. Going to close this for now.

Thanks for the effort.

PR #195, opened 2016-10-11
Added home page link on nav bar

cfmitrah:


DominicWatson:
This doesn't feel right to me. Homepage is on a different nav level so auto including it doesn't feel correct. Also, it is a simple matter to code this into any nav menu should it be required. Going to close this but please re-open if you feel particularly strongly about it.

PR #196, opened 2016-10-11
Fix migration version detection

jjannek:
DB migration fails if using a git repository clone as the version is not properly detected (missing generated version.json file). This PR extends the service to also look at the version in the box.json file and in a git branch name. It should be backwards compatible as the current logic wasn't changed beside the db migration case.


DominicWatson:
Perfect, thanks @jjannek. This has been merged into the upcoming release.

PR #197, opened 2016-10-13
Feature presidecms 576 asset manager issue

tommyw1314:
fix error on duplicated file exist in target folder


DominicWatson (changes requested):
_Deleting this review to try again properly_ (nothing to see here)

DominicWatson (approved):

system/services/assetManager/AssetManagerService.cfc
```diff
@@ -210,13 +211,16 @@ component displayName="AssetManager Service" {
 		  required string  type
 		, required string  size
 		, required string  folderId
+		,          string  assetId
```
DominicWatson: This argument should either be required or have a default value (""). I think it should probably be required.

system/services/assetManager/AssetManagerService.cfc
```diff
@@ -210,13 +211,16 @@ component displayName="AssetManager Service" {
...
 		,          string  currentFolderId = ""
 		,          boolean throwIfNot   = false
 		,          struct  restrictions = getFolderRestrictions( arguments.folderId )
 	) {
 		var typeDisallowed = restrictions.allowedExtensions.len() && !ListFindNoCase( restrictions.allowedExtensions, "." & arguments.type );
 		var sizeInMb       = arguments.size / 1048576;
 		var tooBig         = restrictions.maxFileSize && sizeInMb > restrictions.maxFileSize;
+		var fileExist      = _getAssetDao().dataExists( filter = { id = arguments.assetId } );
```
DominicWatson: What we should be checking for here is that _another_ asset does not already exist in the new folder that has the same _name_ as the current asset.

If we just check for the existance of the asset, then we will never be able to move assets between folders.

DominicWatson (changes requested):
I believe there are some issues around the logic used to check uniqueness of assets within a folder. Please see review comments.

tommyw1314:
i have removed the assetId and do checking based on the asset_folder & title.

DominicWatson:
Great, thanks @tommyw1314. Merging into the upcoming release.

PR #198, opened 2016-10-13
Feature presidecms 585 change user agent dbtype

john5onCheng:


DominicWatson:
I think this is probably directed at the wrong branch (hence many file changes, etc.). Going to reject this. However, I have manually made the change directly in source. Thanks @john5onCheng

PR #199, opened 2016-10-15
PRESIDECMS-605 Postgres issues with 10.7 update

sebduggan:
Fix to update script and column alter code


DominicWatson:
Fantastic, thanks for this. Will be released with 10.7.3

PR #200, opened 2016-10-19
PRESIDECMS-93: Added test connection button and functionality on emai…

cfmitrah:
…l setting page


DominicWatson:
Thanks for this @cfmitrah. We've moved all this to "Email centre" though with a validation logic built into the SMTP provider.

Closing.

PR #201, opened 2016-10-21
Fixed the issue on mime type calculation in pdf derivatives.

cfmitrah:


DominicWatson:
Superb, thanks @cfmitrah

PR #202, opened 2016-10-21
PRESIDECMS-610 Don’t use “user” as a table alias!

sebduggan:
Easier to change the alias to security_user than to make it automatically escape all select fields…


DominicWatson:
Fantastic thanks for this. Will be released with 10.7.3

PR #203, opened 2016-10-24
PRESIDECMS-345 : Make user group as required for admin users.

cfmitrah:


system/forms/preside-objects/security_user/admin.edit.self.xml
```diff
@@ -10,6 +10,7 @@ This form is used for the "edit user" form in the user admin section of the admi
 	<tab id="basic">
 		<fieldset id="basic">
 			<field name="active" deleted="true" />
+			<field binding="security_user.groups" required="false"/>
```
DominicWatson: A user should not be able to edit their own groups :) Please remove this line.

DominicWatson (changes requested):
Looking good, but please see review comment.

system/forms/preside-objects/security_user/admin.edit.self.xml
cfmitrah: Removed edit functionality for user group

DominicWatson:
Nice one, thanks for this. Merged into the upcoming 10.8.0 release.

PR #204, opened 2016-10-27
PRESIDECMS-611 Fix aggregate queries for Postgres

sebduggan:
This fixes the issue for PostgreSQL by duplicating the group by logic from the MSSQL adapter.

However, longer term it would be better to make explicit group by clauses part of the core, as it is also the default setting for MySQL 5.7 and later. And there might be a slightly neater way of handling it if it's baked in at a lower level...


DominicWatson:
This is great, thanks @sebduggan. Maybe you'll convert us to Postgres ;)

PR #206, opened 2016-11-09
PRESIDECMS-245 : Disable the iFrame in CKEditor by default.

cfmitrah:


DominicWatson:
Nice one, thanks @cfmitrah. Merged into upcoming 10.8.0 release

PR #207, opened 2016-11-09
PRESIDECMS-606 : After adding new record, page redirects to correspon…

cfmitrah:
…ding edit_record page in datamanager.


DominicWatson:
Thanks for this Selva. Putting this on hold though for now. The idea behind the ticket was to have some kind of _option_ for where the system takes you after adding a record. Let's revisit later.

PR #208, opened 2016-11-15
PRESIDECMS-639 : Fixed the 404 page issue in generating rulesEngine c…

cfmitrah:
…onditions.


DominicWatson:
I think this could do with individual arguments for each of the redirect rules. We should be able to supply the full URL of each redirect, rather than just the action. The `_editRecordAction()` method already does this and we could use that as an example.

cfmitrah:
- Updated addConditionAction() in rulesEngine.cfc to redirect when `I want to add another Rules engine condition` is enabled.
- Passed queryString as a full URL to redirect.

DominicWatson:
Great, thanks guys.

Merged into upcoming release.

PR #209, opened 2016-11-16
PRESIDECMS-626 : Added the message column on error log listing page.

cfmitrah:


system/services/errors/ErrorLogService.cfc
```diff
@@ -54,7 +54,10 @@ component displayName="Error Log Service" {
 		var errors = [];
 
 		for( var file in files ) {
-			errors.append( { date=file.dateLastModified, filename=file.name } );
+			var errorMessage   = REMatchNoCase( "<\s*title[^>]*>(.*?)<\s*\/\s*title>", FileRead( _getLogDirectory() & "/" & file.name ) );
+			var messageContent = ReplaceList( errorMessage[1] , "<title>,</title>", "#chr(10)# ,#chr(10)#" );
```
DominicWatson: We ought to be a little careful here. What if, for whatever reason, `REMatchNoCase()` returns an empty array here? 

Perhaps: 
```
var messageContent = errorMessage.len() ? ReplaceList( errorMessage[1] , "<title>,</title>", "#chr(10)# ,#chr(10)#" ) : "";
```

system/views/admin/errorLogs/index.cfm
```diff
@@ -34,6 +35,9 @@
 							</a>
 						</td>
 						<td>
+							#left( log.message, "30" )#...<span class="fa fa-caret-down" data-rel="popover" data-trigger="hover" data-placement="left" data-content="<pre>#log.message#</pre>" data-html="true"></span>
```
DominicWatson: We need to make sure that we escape the log message when inserting as an attribute in HTML:

```
... data-content="#HtmlEditFormat( '<pre>#log.message#</pre>' )#" ...
```
```

DominicWatson (changes requested):
Looking good Selva, thank you. Please see the review comments - a couple of minor points.

system/services/errors/ErrorLogService.cfc
cfmitrah: Updated messageContent for avoid empty array.

system/views/admin/errorLogs/index.cfm
cfmitrah: Used HTML attribute for log message.

DominicWatson:
Nice one, thanks @cfmitrah. Merging into upcoming release (10.8.0)

DominicWatson (approved):

PR #211, opened 2016-11-25
PRESIDECMS-651: Implemented clone form functionalities.

cfmitrah:


system/services/formbuilder/FormBuilderService.cfc
```diff
@@ -959,6 +959,46 @@ component {
...
+		, required string name
+		, required string description
+	) {
+		var originalFormData = getForm( id=arguments.basedOnFormId );
+		var cloneFormData = structNew();
+
+		for( var column in originalFormData.columnList ) {
+			if( !listFindNoCase( "id,datecreated,datemodified,_version_is_draft,_version_has_drafts", column ) ) {
+				if( listFindNoCase( "name,description", column ) ) {
+					cloneFormData[ "#column#" ] = arguments[ "#column#" ];
```
DominicWatson: `cloneFormData[ "#column#" ]` should just be `cloneFormData[ column ]` (same for `arguments[ column ]`). See also lines below.

system/views/admin/formbuilder/cloneForm.cfm
```diff
@@ -0,0 +1,28 @@
+<cfscript>
+	var id = rc.id ?: "";
+	var formId = createUUID();
```
DominicWatson: This has the potential to create invalid HTML element IDs (they must not start with a number). I think ok to use "formbuilder-clone-form" or somesuch here.

system/services/formbuilder/FormBuilderService.cfc
```diff
@@ -959,6 +959,46 @@ component {
 		}
 	}
 
+	public string function cloneForm(
+		  required string basedOnFormId
+		, required string name
+		, required string description
+	) {
+		var originalFormData = getForm( id=arguments.basedOnFormId );
+		var cloneFormData = structNew();
```
DominicWatson: Here, I think this is more accurate:

```
var cloneFormData = { name=arguments.name, description=arguments.description };
```

Then further down, when you loop over the `originalFormData.columnList`, you can ignore these fields and remove the logic on line 972-3.

system/services/formbuilder/FormBuilderService.cfc
```diff
@@ -959,6 +959,46 @@ component {
...
+
+		for( var column in originalFormData.columnList ) {
+			if( !listFindNoCase( "id,datecreated,datemodified,_version_is_draft,_version_has_drafts", column ) ) {
+				if( listFindNoCase( "name,description", column ) ) {
```
DominicWatson: See comment above

DominicWatson (changes requested):
Hey Mani, this is looking really great, thank you. Please see the review comments - a couple of minor tweaks, mostly just tidying.

Thanks!

cfmitrah:
@DominicWatson  
	* Created valid HTML element IDs, using "formbuilder-clone-form".
	* Simplified the code while looping 'originalFormData.columnList'

system/views/admin/formbuilder/cloneForm.cfm
cfmitrah: Created valid HTML element IDs, using "formbuilder-clone-form".

system/services/formbuilder/FormBuilderService.cfc
cfmitrah: Simplified the code while looping 'originalFormData.columnList' by 
<code>var cloneFormData = { name=arguments.name, description=arguments.description };</code>

DominicWatson:
Super, thanks @cfmitrah. Merged into upcoming release.

PR #212, opened 2016-11-29
PRESIDECMS-656

john5onCheng:
Ensures object picker is able to filter from related simple value field.


john5onCheng:
Hi Dom, I've revert previous changes and apply the recommended approach.

DominicWatson:
This looks about right - I tried it out locally though and couldn't get it to work. There are a few things here that I'd refactor too. Let's take a look together as soon as reasonable.

john5onCheng:
Hi Dom, I've found the error that you mentioned and have get it fixed. Please let me know which line you would like to refactor.

system/handlers/admin/DataManager.cfc
```diff
@@ -194,11 +194,24 @@
 
 			_checkPermission( argumentCollection=arguments, key="read", object=objectName );
 
+			var filterByField = rc.filterByField ?: "";
+			var extraFilters  = [];
+
+			loop list=filterByField index="indexField"{
```
DominicWatson: `indexField` is not `var'd` here and I would avoid looping a list - especially using cfloop syntax. Prefer this:

```
var filterByFields = ListToArray( rc.filterByField ?: "" );
for( var filterByField in filterByFields ) {
    // ...
}
```

system/assets/js/admin/presidecore/preside.uber.select.js
```diff
@@ -365,10 +365,21 @@
 				if ( this.local_options[i].empty ) { this.local_options.splice( i, 1 ); }
 			}
 
+			var prefetch_url = this.form_field.getAttribute( "data-prefetch-url" );
+			var remote_url   = this.form_field.getAttribute( "data-remote-url" );
```
DominicWatson: in our *js*, we'd like `var` definitions at the _top_ of functions and using the style:

```
var variable1 = true
  , variable2 = false
  , variable3, variable 4; // not yet set vars
```

These vars should be moved up aroud line 355 (and declared in the list style).

system/assets/js/admin/presidecore/preside.uber.select.js
```diff
@@ -500,7 +510,31 @@
...
+					$uberSelect.setup_search_engine();
+				});
+			}
+		};
+
+		UberSelect.prototype.setup_filter = function() {
+
+			this.filter         = "" ;
+
+			var filterBy        = this.form_field.getAttribute( "data-filter-by" );
```
DominicWatson: All the `var` definitions should be moved to the top of the function here:

```
var filterBy        = this.form_field.getAttribute( "data-filter-by" )
  , filterByField   = this.form_field.getAttribute( "data-filter-by-field" )
  , filterByValue;

this.filter       = "";
this.filter_field = $( "input[name='" + filterBy + "']" );

filterByValue = this.filter_field.val();
if ( typeof filterByValue !== "undefined" ) {
// ...
}
```

DominicWatson (changes requested):
Looking good, please see review comments re points that need refactoring.

john5onCheng:
hi Dom, requested changes has been made

DominicWatson (approved):

DominicWatson:
Perfect, testing now.

PR #214, opened 2016-12-06
PRESIDECMS-661: Created Clone question on FormBuilder.

cfmitrah:


DominicWatson:
Fantastic, thanks @cfmitrah. Merged into upcoming 10.8.0 release.

PR #219, opened 2016-12-10
ACF_echo: Changed writeOutput() method instead of echo()

cfmitrah:


DominicWatson:
Fantastic, thanks guys. Merged into 10.8.0, pending release.

PR #220, opened 2016-12-10
ACF_elseif: The elseif syntax replaced by else if

cfmitrah:


DominicWatson:
Fantastic, thanks guys. This has been merged into upcoming 10.8.0. Pending release.

PR #222, opened 2016-12-13
Use DateCompare() to compare dates

vivienwong:


DominicWatson:
That's great, thanks @vivienwong. Merged into upcoming 10.8.0. Pending release.

PR #223, opened 2016-12-13
PRESIDECMS-583: Displayed a thumbnail while hover on selected asset p…

cfmitrah:
…icker image.


DominicWatson:
Nice one, merged into upcoming release. Thanks @cfmitrah

PR #224, opened 2016-12-14
PRESIDECMS-671 : Added image dimensions on image picker dropdown

cfmitrah:


DominicWatson:
Great, thanks @cfmitrah. Merged into upcoming release (with some efficiency tweaks: width and height are stored against the asset record so no need to recalculate every request).

PR #226, opened 2016-12-16
ACF_semicolon: Fixed semicolon issue on system and support too.

cfmitrah:


DominicWatson:
Great stuff, thanks @cfmitrah. Merging into upcoming 10.8.0 release

PR #231, opened 2016-12-23
PRESIDECMS-682 : Updated bootstrap.cfc to get customTag paths from ac…

cfmitrah:
…tive extensions.


DominicWatson:
This is great, thanks Selva. Merging into 10.8.0

DominicWatson:
This breaks for me and isn't quite right. We shouldn't be invoking the extensions service on every request. Closing this PR for now in wait of a better approach. I tried refactoring but nothing doing. Will need more time to figure out the right approach.

PR #232, opened 2016-12-23
PRESIDECMS-683

john5onCheng:
Show the edit button when selected single title is too long


DominicWatson:
Fantastic, thanks for this @john5onCheng (and thanks for the super clear jira ticket).

PR #233, opened 2016-12-30
Feature presidecms 684 data manager sort order

nelsonchuah:
Add function to update orderBy to use label when order by field is an object, previously it was using object's id to sort when it's an object


DominicWatson:
Brilliant, thanks for this fix @nelsonchuah

PR #234, opened 2017-01-03
PRESIDECMS-586

john5onCheng:
Ensures there is an option to force resize on quality


DominicWatson:
Great, thanks for this fix @john5onCheng.

PR #239, opened 2017-01-11
PRESIDECMS-693 Page Expiry date/Embargo date doesn't filter on the time

seonpks:
removed the setting.now variable and change into mysql now() function

### DIFF
--- a/system/config/Config.cfc
+++ b/system/config/Config.cfc
@@ -309,8 +309,11 @@ component {
 		};
 
 		settings.filters = {
-			livePages = {
-				filter = "page.trashed = '0' and page.active = '1' and ( page.embargo_date is null or NOW() > page.embargo_date ) and ( page.expiry_date is null or NOW() < page.expiry_date )"
+			livePages = function(){
+				return {
+						 	  filter       = "page.trashed = '0' and page.active = '1' and ( page.embargo_date is null or :now > page.embargo_date ) and ( page.expiry_date is null or :now < page.expiry_date )"
+							, filterParams = { "now" = { type="cf_sql_timestamp", value=Now() } }
+						};
 			}
 			, activeFormbuilderForms = { filter = { "formbuilder_form.active" = true } }
 		};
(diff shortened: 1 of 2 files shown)


system/config/Config.cfc
```diff
@@ -310,8 +310,7 @@ component {
 
 		settings.filters = {
 			livePages = {
-				  filter       = "page.trashed = '0' and page.active = '1' and ( page.embargo_date is null or :now > page.embargo_date ) and ( page.expiry_date is null or :now < page.expiry_date )"
-				, filterParams = { "now" = { type="cf_sql_date", value=Now() } }
+				filter = "page.trashed = '0' and page.active = '1' and ( page.embargo_date is null or NOW() > page.embargo_date ) and ( page.expiry_date is null or NOW() < page.expiry_date )"
```
DominicWatson: The only issue with using raw Now() in the SQL is database engine compatibility. There is a database adapter method for getting the 'now' function for the given database engine being used, but that is not possible to access in a static filter.

I think the real fix here is the param type. I think it needs to be something like `cf_sql_timestamp` rather than `cf_sql_date`. Can you try that?

system/services/siteTree/SiteTreeService.cfc
```diff
@@ -509,6 +509,7 @@ component {
 				, savedFilters       = savedFilters
 				, orderBy            = "sort_order"
 				, allowDraftVersions = args.allowDrafts
+				, useCache           = false
```
DominicWatson: I'd rather use cache here if we can. The cache timeout is 20 minutes, so we are talking about being accurate to 20 mins for the embargo/expiry date. Given that the navigation is invoked on every page, it makes sense to cache it by default.

DominicWatson (changes requested):
Hey @seonpks thanks, for the PR :). Please see review comments.

seonpks:
Hi dom, i've done the changes based on your review. please have a check.

DominicWatson:
Nice one, thanks @seonpks

PR #240, opened 2017-01-12
Feature presidecms 695 asset url not updated when new version added

seakchiew:
Check the existing asset_url if it contains the latest version before returning the url, else generate a new url so that new version is always honoured.


DominicWatson:
Thanks for this @seakchiew - can you check the tests on this though? looks like its breaking a couple of tests. Would be good to update the tests to ensure we're testing for the right thing here. 

Also, we should probably change the code that generates the URL so that it _always_ includes a version in the first place?

PR #241, opened 2017-01-12
PRESIDECMS-692

john5onCheng:
Ensures ajax is running on async: false


john5onCheng:
please cancel this PR

PR #242, opened 2017-01-12
PRESIDECSM-692

john5onCheng:
Ensures ajax connection for chosen select is running on async = false


DominicWatson:
To make things easier with asset compiling, etc. I'm going to reject this PR but make the changes manually (so thanks for the investigation and fix). 

I've also created a separate JIRA issue - we can not and should not work on already released tickets.

PR #243, opened 2017-01-13
PRESIDECMS-695 update asset_url when version is made active.

seakchiew:
I have made so that the asset active_url is updated when an asset_version is made active. This should cover addAsset and makeVersionActive action. URL is generated based on the getAssetUrl method.


DominicWatson:
Nice one, thanks Seak Chiew. Released as its own hotfix: 10.7.25

PR #245, opened 2017-01-17
PRESIDECMS-707

john5onCheng:
Ensures OG meta tag have property attribute


DominicWatson:
I've manually fixed this issue as unsure about a bit of this pull request (so closing the PR).

What is the addition of the "label" page property? This isn't a core preside page property so I don't want to pull that change.

Thanks for pointing out the core fix though. If you could explain the second change, perhaps it could make its own PR (or maybe it belongs in your specific application only)?

PR #247, opened 2017-01-19
PRESIDECMS-714 Add URL args to _addRecordAction()

sebduggan:
One difference between add and edit actions is that some redirects for add require the newly-generated id as a parameter.

The placeholder {newid} in any of these URLs will be replaced by the generated id.


DominicWatson:
Nice one, thanks @sebduggan - merged into upcoming 10.8.0 release.

PR #250, opened 2017-01-26
PRESIDECMS-732 - Task manager log date time not formatted

nodoherty:
Implemented some date formatting to the date time display of the log so it's less system like


nodoherty:
Going to point the PR to Stable branch instead of the 10.8 release

PR #252, opened 2017-01-27
Feature presidecms 735 bug missing paramater adapter

nodoherty:
Applied a fix for the missing adapter argument in the interceptor explicitly without having to depend on one being passed

### DIFF
--- a/system/interceptors/MultilingualPresideObjectInterceptor.cfc
+++ b/system/interceptors/MultilingualPresideObjectInterceptor.cfc
@@ -1,12 +1,11 @@
-component extends="coldbox.system.Interceptor" output=false {
+component extends="coldbox.system.Interceptor" {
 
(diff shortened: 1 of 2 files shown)


DominicWatson:
I'd move this to the service and not the interceptor:

```
public void function mixinTranslationSpecificSelectLogicToSelectDataCall( required string objectName, required array selectFields ) {
    var adapter = $getPresideObjectService().getDbAdapterForObject( arguments.objectName );
    ...
```

The interceptor is then, rightly, just a dump proxy and isn't adding fields to the intercept data that don't need to be there.

nodoherty:
Feedback applied - moved adapter call to inside the service and removed as an argument + removed redundant output directives

DominicWatson:
Nice one, thanks @nodoherty. Merged into the upcoming releases.

PR #255, opened 2017-01-31
PRESIDECMS-747  Don't force lowercase for i18n Translation

eppinger:
LCase( object.... ) replaced in translateResource calls


DominicWatson:
Yikes, there's a lot of them! Nice work, thanks for this. Will pull asap.

DominicWatson:
Merged into upcoming 10.8.0 release.

PR #260, opened 2017-02-21
PRESIDECMS-766 Check presence of labelfield before dataSelect / acces…

vivienwong:
PRESIDECMS-766 Check presence of labelfield in the object or replace with id


DominicWatson:
Nice one, thanks @vivienwong. Merged into upcoming 10.8.0 release.

PR #261, opened 2017-02-22
Feature preside 767 update richeditor quickadd js

nelsonchuah:
Hi Dom,

Have update quickAddForm's resetForm function to manually clear content for richeditor

Thank you
Nelson


system/assets/js/admin/specific/datamanager/quickAddForm/quickAddFormHandler.js
```diff
@@ -72,6 +72,9 @@
 
 	resetForm = function(){
 		$quickAddForm.trigger( "reset" );
+		for (instance in CKEDITOR.instances){
```
DominicWatson: what if `CKEDITOR.instances` does not exist (i.e. there are no richeditors in the form), will this produce a js error (I suspect so)? Also, is there a possibility that there are ckeditor instances outside of the quick add form? I'm guessing *not* as I believe the form is in an iframe but worth checking for too. Please add some defensive logic for that.

Finally, it would make sense to refactor this out to its own mthod:

```
clearRicheditors( $quickAddForm );
```

DominicWatson (changes requested):
Thanks for this @nelsonchuah - please see review comments.

nelsonchuah:
Hi Dom,

Have updated code,
1) add checking for CKEDITOR ( CKEDITOR's instances will always be available in CKEDITOR exist )
- richeditor outside of quickadd form wouldn't be affected (have tested this)
2) refractor out code

Thank you
Nelson

system/assets/js/admin/specific/datamanager/quickAddForm/quickAddFormHandler.js
```diff
@@ -72,11 +72,17 @@
 
 	resetForm = function(){
 		$quickAddForm.trigger( "reset" );
-		for (instance in CKEDITOR.instances){
-			CKEDITOR.instances[instance].setData("");
-		}
+		clearRicheditor();
 	};
 
+	clearRicheditor = function() {
```
DominicWatson: `clearRicheditor` is not `var` scoped. It should be added to the list of `var` scoped variables at the top of the function (line 4).

DominicWatson (changes requested):
Looking great, thanks @nelsonchuah. One little tweak left :)

nelsonchuah:
added function into var scope now :)

DominicWatson (approved):

DominicWatson:
Perfect, thanks @nelsonchuah

PR #262, opened 2017-02-22
PRESIDECMS-608 One-to-many configurator form control

sebduggan:


system/handlers/formcontrols/OneToManyConfigurator.cfc
```diff
@@ -0,0 +1,41 @@
...
+		var targetFk      = args.relationshipKey ?: args.sourceObject;
+		var useVersioning = Val( rc.version ?: "" ) && presideObjectService.objectIsVersioned( targetObject );
+		var hasSortOrder  = presideObjectService.getObjectProperties( targetObject ).keyExists( "sort_order" );
+		var orderBy       = hasSortOrder ? "sort_order" : "";
+		
+		if ( Len( Trim( args.savedData.id ?: "" ) ) ) {
+			var records       = presideObjectService.selectData(
+				  objectName       = targetObject
+				, filter           = { "#targetFk#"=args.savedData.id }
+				, selectFields     = [ "id", "label" ]
```
DominicWatson: what happens here when the target object uses a label field other than "label", or doesn't have a label field at all? We _can_ use:

```
[ "id", "${labelfield} as label" ]
```

But perhaps it makes sense just to select all fields by not supplying a select fields array at all?

system/assets/js/admin/specific/datamanager/configuratorForm/configuratorFormHandler.js
```diff
@@ -14,25 +14,9 @@
...
-						if ( inputValues[ i ] == item.value ) {
-							return item.text;
-						}
-					}
-				} ).get().join( ', ' );
-			});
-
+			// $configuratorForm.find(':checkbox:not(:checked)').attr('value', '0').prop('checked', true);
+			var formData = $configuratorForm.serializeObject();
+			console.log(formData);
```
DominicWatson: Some debugging left over ;)

system/services/presideObjects/PresideObjectService.cfc
```diff
@@ -1081,6 +1083,14 @@ component displayName="Preside Object Service" {
 		);
 
 		for (var record in records ) {
+			var formName = record.configurator__formname ?: "";
```
DominicWatson: I'd like to remove this logic if possible. Is the purpose to ensure that checkbox boolean fields have a present variable? I think this logic is the responsibility of the code that prepares the json (i.e. all the required fields should be present and its not the responsibility of the preside object service to start validating the input).

I think this should be safe to remove altogether (along with the line that gets the forms service). Or is this logic absolutely required?

DominicWatson (changes requested):
Looks great but for the forms service injection which we can discuss later if needed (though I think is safe to remove).

Also a bit of debugging left over in the js :)

system/assets/js/admin/specific/datamanager/configuratorForm/configuratorFormHandler.js
sebduggan: Oops...

system/services/presideObjects/PresideObjectService.cfc
sebduggan: You're right, it was just there to ensure unchecked checkboxes were recorded. Now implemented in the Javascript.

DominicWatson:
Nice one, thanks @sebduggan. Merging into upcoming 10.8 :)

PR #263, opened 2017-02-28
PRESIDECMS-779 datepicker revalidation

sebduggan:
Also includes CSS fix to make sure hidden text is off-screen even on large displays


DominicWatson:
Nice one, thanks @sebduggan. Merged into upcoming 10.8.0 release.

PR #268, opened 2017-03-04
PRESIDECMS-783 generators on versioned objects

sebduggan:


DominicWatson:
Nice one, thanks @sebduggan. Merging now.

PR #271, opened 2017-03-08
PRESIDECMS-789 Make ObjectPicker filters work with non-ajax and quickadd

sebduggan:
- Object picker filters are passed through to quickAdd forms
- Hidden fields are appended to a form's savedData record, for access by all form controls. This is also now passed through to add forms
- Non-AJAX object pickers can be filtered based on values in this extended savedData
- A non-multiple object picker select menu can now be marked as non-deselectable, which will prevent the value from being cleared. This is especially useful if another control is filtering on the value of this control.


DominicWatson (approved):
Perfect. Thanks @sebduggan

PR #274, opened 2017-03-11
PRESIDECMS-791 - Adding missing i18n for Url Redirects when adding a rule

nodoherty:
Minor update to the core cms.properties i18n file...


DominicWatson:
Nice one, thanks @nodoherty

PR #277, opened 2017-03-19
PRESIDECMS-727 Count-only queries shouldn’t have order by in MSSQL

sebduggan:


DominicWatson:
Nice one, thanks @sebduggan

PR #278, opened 2017-03-20
PRESIDECMS-751 Add renderer arg to readonly form control

sebduggan:
Date and datetime auto-select the appropriate renderer

### DIFF
--- a/system/handlers/formcontrols/ReadOnly.cfc
+++ b/system/handlers/formcontrols/ReadOnly.cfc
@@ -10,10 +10,6 @@ component {
 		args.renderer  = args.renderer ?: contentRendererService.getRendererForField( prop );
 
 		if ( !contentRendererService.rendererExists( args.renderer ) ) {
-			args.renderer = type;
-		}
-
-		if ( renderer == "none" ) {
 			args.renderer = "";
 		}
 
--- a/system/handlers/formcontrols/ReadOnly.cfc
+++ b/system/handlers/formcontrols/ReadOnly.cfc
@@ -1,0 +1,19 @@
+component {
+
+	property name="presideObjectService" inject="presideObjectService";
+
+	public string function index( event, rc, prc, args={} ) {
+		var renderer = args.renderer ?: "";
+		var type     = presideObjectService.getObjectPropertyAttribute( args.sourceObject, args.name, "type" );
+
+		if ( listFindNoCase( "date,datetime", type ) ) {
+			args.renderer = type;
+		}
+
+		if ( renderer == "none" ) {
+			args.renderer = "";
+		}
+
+		return renderView( view="formcontrols/readOnly/index", args=args );
+	}
+}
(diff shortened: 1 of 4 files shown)


system/handlers/formcontrols/ReadOnly.cfc
```diff
@@ -0,0 +1,19 @@
+component {
+
+	property name="presideObjectService" inject="presideObjectService";
+
+	public string function index( event, rc, prc, args={} ) {
+		var renderer = args.renderer ?: "";
+		var type     = presideObjectService.getObjectPropertyAttribute( args.sourceObject, args.name, "type" );
```
DominicWatson: See `/system/services/rendering/ContentRendererService.cfc$getRendererForField()`. This would be a better start and re-use existing logic in terms of finding the best default renderer for the field.

DominicWatson (changes requested):

system/views/formcontrols/readonly/index.cfm
```diff
@@ -1 +1,11 @@
-<cfoutput><span class="read-only">#( HTMLEditFormat( args.defaultValue ?: ( args.savedValue ?: "" ) ) )#</span></cfoutput>
\ No newline at end of file
+<cfscript>
+	renderer = args.renderer     ?: "";
+	context  = args.context      ?: "default";
```
DominicWatson: use "readonly" context + make `args.rendererContext` to avoid conflicting "contexts" (forms can be rendered in a context)

DominicWatson (changes requested):

system/handlers/formcontrols/ReadOnly.cfc
```diff
@@ -0,0 +1,19 @@
...
+
+		if ( listFindNoCase( "date,datetime", type ) ) {
+			args.renderer = type;
+		}
+
+		if ( renderer == "none" ) {
```
DominicWatson: we should use `if ( !contentRendererService.rendererExists( renderer ) ) {...}`;

system/handlers/formcontrols/ReadOnly.cfc
```diff
@@ -0,0 +1,22 @@
...
+	property name="contentRendererService" inject="contentRendererService";
+
+	public string function index( event, rc, prc, args={} ) {
+		var properties = presideObjectService.getObjectProperties( args.sourceObject );
+		var prop       = properties[ args.name ] ?: {};
+		
+		args.renderer  = args.renderer ?: contentRendererService.getRendererForField( prop );
+
+		if ( !contentRendererService.rendererExists( args.renderer ) ) {
+			args.renderer = type;
```
DominicWatson: `type` is not defined I think? I think this should be `""`? Can then also ditch the `renderer == "none"` check below.

DominicWatson (changes requested):

DominicWatson:
Brill, merging now.

DominicWatson (approved):

PR #279, opened 2017-03-22
PRESIDECMS-810: added ondelete and onupdate annotations

jjannek:
to cascase-if-no-cycle-check to solve MSSQL DB sync problems


DominicWatson:
Fantastic, thanks for this @jjannek. Merging now.

PR #280, opened 2017-03-24
Feature presidecms 820 formbuilder content field error

nelsonchuah:
Hi Dom,

Content field is throwing error when editing as it don't have name and label, have added a default value to it.

Affected version is 10.8 only. Code doesn't exist in 10.7

Thank you
Nelson


DominicWatson:
Fantastic, thanks @nelsonchuah.

PR #282, opened 2017-03-24
PRESIDECMS-607 Label renderer additions

sebduggan:
Add group by and implement caching of label renderers (label renderer handler is now included in the cachebuster for object pickers)


DominicWatson:
Nice one, merging. Thanks Seb.

PR #284, opened 2017-03-27
PRESIDECMS-824: Error occurs if trying to add a record via the one-to-many manager form control

jjannek:
added missing semicolon in one-to-many record form


DominicWatson:
Nice one, thanks @jjannek

PR #288, opened 2017-03-28
PRESIDECMS-830 Filtering object pickers wait for pickers they depend on

sebduggan:


system/handlers/admin/DataManager.cfc
```diff
@@ -200,11 +200,19 @@
...
 
 			for( var filterByField in filterByFields ) {
-				if( !isEmpty( rc[filterByField] ?: "" ) ){
-					extraFilters.append({
-						  filter       = "#filterByField# = :#filterByField#"
-						, filterParams = { "#filterByField#" = rc[filterByField] }
-					})
+				filterValue = rc[filterByField] ?: "";
+				if( !isEmpty( filterValue ) ){
+					if ( isValid( "UUID", listFirst( filterValue ) ) ) {
```
DominicWatson: Can we be sure that we only want to behave as an array list when first item is a UUID? Probably fine most of the time, but at some point we may well have natural keys for PK (e.g. system generated data like system emails in email centre have PKs that are human readable from config).

What's the reasoning behind this check?

system/assets/js/admin/presidecore/preside.uber.select.js
```diff
@@ -529,17 +531,22 @@
...
+				filterBy      = filterBy.split( ',' );
+				filterByField = filterByField.split( ',' );
 
-				if ( filterInput.length ) {
-					this.filter_field = filterInput;
-					filterByValue = this.filter_field.val();
-				} else {
-					filterByValue = cfrequest[ filterBy ] || null;
-				}
+				for( var i=0; i<filterBy.length; i++ ) {
```
DominicWatson: Define `var` variables at the top of function definition with js (see Pixl8 styleguide)

DominicWatson (changes requested):

DominicWatson:
Nice one, thanks Seb. Testing locally now.

PR #289, opened 2017-03-28
Feature PRESIDECMS-832 - Update read only form control to handle when no source object is available

nodoherty:
NOTE: subtle difference when calling `renderForm()` and `renderFormControlForObjectField()`


DominicWatson:
Nice one, thanks @nodoherty. Merged.

PR #290, opened 2017-03-29
Feature presidecms 828 add edit page interceptor

wotzhs:
Fix issue with announceInterception not using preside superclass


DominicWatson:
Nice one, thanks @Wotzhs

PR #291, opened 2017-03-29
PRESIDECMS-830 obj picker filter by - fix multiple filters

sebduggan:


DominicWatson:
Nice one, thanks @sebduggan

PR #292, opened 2017-03-30
Feature presidecms 836 formbuilder email template view erroring

nelsonchuah:
Hi Dom,

Have set default value for **args.noResponse** when _preRenderResponses() is called from other functions i.e. (htmlEmail, textEmail)

Thank you
Nelson


DominicWatson:
Nice one, thanks @nelsonchuah

PR #294, opened 2017-03-30
PRESIDECMS-799

john5onCheng:
Increase Uri and user agent max length


DominicWatson:
Brilliant, thanks @john5onCheng. Merged into the upcoming 10.8.0 release.

PR #297, opened 2017-03-31
PRESIDECMS-844

john5onCheng:
Correct the parameter


DominicWatson:
Nice one, thanks @john5onCheng

PR #298, opened 2017-03-31
PRESIDECMS-841 error on is array by null value

seonpks:
fixed the error from isArray when param.value is NULL


DominicWatson:
Great, thanks @seonpks - merging this now.

PR #299, opened 2017-04-02
Update page.md

alexskinner:
Fix for a typo thanks Andrew Jackson


DominicWatson:
Nice one, thanks. Will get released with 10.8.0

PR #301, opened 2017-04-06
Feature presidecms 851 fix frontend editing for richeditor

nelsonchuah:
Hi dom,

Have updated setVersionContent, to set content for richeditor. Wasn't updating previously

Thank you
Nelson


sebduggan:
Thanks Nelson :-)

PR #302, opened 2017-04-10
PRESIDECMS-852

john5onCheng:
Patch the UUID not generated for translation versioning table


sebduggan:
Thanks @john5onCheng. Merged into the upcoming 10.8.0 release.

PR #304, opened 2017-04-12
Feature PRESIDECMS-858 - Fix the time on the task manager once run time starts evaluating minutes hours

nodoherty:
Using the integer division operator "\" rather than the standard division operator "/" we get the integer result back for minutes which is then NOT rounded up and causing inaccuracy in the displayed time taken


nodoherty:
Looks like some other commits leaked into mines...

PR #305, opened 2017-04-12
Feature PRESIDECMS-858 - Fix the time on the task manager once run time starts evaluating minutes

nodoherty:
Switching to the integer division operator means there is no remainder making the time tak calculations more accurate when the seconds accumulate over the 30 second mark for each minute.


nodoherty:
From my original pull request: 
Using the integer division operator "\" rather than the standard division operator "/" we get the integer result back for minutes which is then NOT rounded up and causing inaccuracy in the displayed time taken

DominicWatson:
Nice one, thanks @nodoherty :) Merged into upcoming 10.8.0 release.

PR #308, opened 2017-04-19
PRESIDECMS-862 Make js date validators match server-side

sebduggan:


DominicWatson:
Fantastic, thanks Seb. Merged into upcoming release.

PR #309, opened 2017-04-21
Mssql alternative paging

jjannek:
I created a MS SQL 2012+ DB Adapter that uses the newer paging concept. It works alongside the standard MS SQL Adapter.


DominicWatson:
Apologies, this was closed because the release is finished and the branch no longer exists! Please resubmit and target Stable :)

PR #312, opened 2017-04-27
Mssql alternative paging

jjannek:
Alternative paging for MS SQL 2012 and above.
Sorry for the additional changes as I merged in the 10.9 branch and did not get it out again (only version bumps). Let me know how to fix this if it does not work for you.


DominicWatson:
Awesome, thanks @jjannek :) Getting released with 10.8.3

PR #315, opened 2017-05-04
PRESIDE-876

john5onCheng:
Set default forceJoin to left


DominicWatson:
It really should be "". Forcing joins that should be "inner" to "left" is a bad idea. When "", the system will figure out the correct join to use based on whether or not many-to-one relationship fields are required or not. If the FK is mandatory, it never makes sense for there to be a left join.

PR #316, opened 2017-05-04
PRESIDECMS-756: fixed REST processing

jjannek:
(argument name for default value in requestcontext method getHttpHeader changed


DominicWatson:
Awesome, thanks @jjannek

PR #317, opened 2017-05-08
PRESIDECMS-866

john5onCheng:
Set all exported cell type to string to avoid error rendering to wrong type


DominicWatson:
Nice one, thanks @john5onCheng . Merged and released in 10.8.2.

PR #319, opened 2017-05-12
PRESIDECMS-675: ensuring uniqueness of HTML element IDs 

jjannek:
...for generated radio/checkbox formcontrols

Sorry for the indentation mess. In my editor it showed up as tab-indented and I changed it to indent using 4 spaces (i thought that is used in Preside by default). Therefore the whole indentation changed. :-(

The real changes are:
- added loop index as a suffix to input element IDs for checkbox/radio lists and the matrix control
- renamed the inner loop index variable on the matrix formcontrol ('i' was used twice)


DominicWatson:
Awesome, thanks for the fix @jjannek

PR #320, opened 2017-05-23
PRESIDECMS-888

john5onCheng:
Always allow to download for unknown type document as attachment


DominicWatson:
Great, thanks for this @john5onCheng

PR #321, opened 2017-05-30
PRESIDECMS-540 Change handling of uber select option limits

sebduggan:
Changes made:
- By default, 200 items will be displayed
- This can be set per control using the "displayLimit" attribute
- Setting to 0 will be unlimited - i.e. display all options
- If a control is non-Ajax, it will automatically be unlimited
- Search index for selects will now be tokenised on any non-word character or underscore, as opposed to just spaces as before. This allows searching for e.g. "London" in the string "Europe/London", or "new york" in "America/New_York"


DominicWatson:
Awesome, thanks @sebduggan

PR #323, opened 2017-06-13
PRESIDECMS-909 dateModified select bug for grid fields

sebduggan:


DominicWatson:
Nice one, thanks @sebduggan. Fix coming in 10.8.8 hotfix.

PR #325, opened 2017-06-16
Feature presidecms 921 excel export invalid characters in name too long

nodoherty:
Simple fix to ensure a sheet title doesn't exceed the 31 character limit given the title comes from the objects i18n


DominicWatson:
Fantastic, thanks @nodoherty

PR #327, opened 2017-06-29
PRESIDECMS-934 Fix for HTML labels in disabled object picker

sebduggan:


DominicWatson:
Nice one, will make it into the next hotfix (10.8.17). Thanks @sebduggan

PR #328, opened 2017-07-04
PRESIDECMS-937: Check field type before comparing old with new data

tarekabou-chanab:
Now checking for dbtype date/datetime and using dateCompare instead of Compare


DominicWatson:
Looks perfect, thanks @tarekabou-chanab

PR #330, opened 2017-07-06
PRESIDECMS-941

john5onCheng:
Add full screen option


DominicWatson:
Nice one, thanks @john5onCheng - will make it into upcoming hotfix (10.8.20)

PR #331, opened 2017-07-11
PRESIDECMS-946: Add 'preside' antisamy policy and make it the default…

nickcervantes:
Hi @DominicWatson, I've added the 'preside' antisamy policy and make it the default.


DominicWatson:
Exemplary, thanks @nick-htmlguys !

PR #332, opened 2017-07-14
PRESIDECMS-955 check stickerBundle file prior to processing

vivienwong:
Replace try catch statement with StickerBundle file exists check to reveal any hidden asset definition error


DominicWatson:
Brilliant, thanks @vivienwong

PR #334, opened 2017-07-20
PRESIDECMS-964 - Return the configured default label if no items or "…

nodoherty:
…None" given no default label also


DominicWatson:
Nice one, thanks @nodoherty

PR #344, opened 2017-08-15
PRESIDECMS-897 non existing derivative type should return 404

sebduggan:


DominicWatson:
Nice one, thanks @sebduggan

PR #345, opened 2017-09-01
PRESIDECMS-1010 Object picker search should work with formula label fields

sebduggan:


DominicWatson:
Brilliant, thanks @sebduggan

PR #346, opened 2017-09-04
PRESIDECMS-1012

john5onCheng:
Increase z-index for frontend dialog


DominicWatson:
Fantastic, thanks @john5onCheng

PR #349, opened 2017-09-21
PRESIDECMS-1025

john5onCheng:
Ensures audit log has no error when email centre is enabled


DominicWatson:
Great, thanks for this @john5onCheng

PR #353, opened 2017-09-21
incorrect path

philippPSH:
templates is incorrect. it took 1h to get it ;-)


DominicWatson:
Yikes! thanks for the fix @philippPSH - will merge asap.

PR #354, opened 2017-09-22
PRESIDECMS-1027 Remove incorrect stripping of translated action witho…

vivienwong:
Remove incorrect stripping of action without path


DominicWatson:
Great, thanks for this @vivienwong

PR #355, opened 2017-09-29
PRESIDECMS-1032 One to many configurator sort order issue

sebduggan:
Issue arises from handling of versioning of configurator data. This is now much improved and solves the sort order problem.


DominicWatson:
Great, thanks for this @sebduggan

PR #358, opened 2017-10-04
PRESIDECMS-1035

john5onCheng:
Avoid duplicated column name returned


DominicWatson:
Nice one, thanks for this @john5onCheng

PR #367, opened 2017-10-24
PRESIDECMS-1049 Fix microsite URL issue

sebduggan:
Revert the code change from PRESIDECMS-905, and instead *force* a lookup of the site domain when performing a previewPage redirect.


sebduggan:
Merged.

PR #369, opened 2017-10-25
PRESIDECMS-1050 live site reload protection

sebduggan:
Preside console requires **--force** flag for **reload all** and **reload db** when in production environment


sebduggan:
Merged.

PR #373, opened 2017-10-27
Update page.md

jjannek:
added missing '@' signs to labelfield and nolabel annotations


DominicWatson:
Nice one, thanks @jjannek - will get updated with the next hotfix.

PR #377, opened 2017-11-03
PRESIDE-1061 enable grey disabled class for the spacer clone link

annbelievable:
Preside-1061
Changes in view/formbuilder/_workbenchFormItem.cfm
If item type does not requiresConfiguration, then the clone link will be unclickable.
Hence the spacer's clone link will not be clickable.


DominicWatson:
Nice one @annbelievable - could you delete the `system/assets/package-lock.json` file though? Looks like that's come through from NPM somehow.

annbelievable:
Hi @DominicWatson,
Kok Ann here, deleted the system/assets/package-lock.json file.
However I'm not sure the solution is a desired one though.

DominicWatson:
Ha, doh - one file out, two in (`.DS_Store`). Could you remove those too?

PR #386, opened 2017-12-06
PRESIDECMS-1078 added MsSql2012 compatible concatenation function to DB adapter

jjannek:


DominicWatson:
Nice one, thanks @jjannek. Merged into 10.9.0

PR #387, opened 2017-12-06
PRESIDECMS-1079 Fix for labelRenderer search & MSSQL select

thomasrotter:


DominicWatson:
Brilliant, thanks @thomasrotter . Merging into upcoming 10.9.0 release so PR stays open until 10.9.0 is final

PR #388, opened 2017-12-06
Feature presidecms 1076 asset manager lucee 5 element at position 1 does not exist in list array index out of range 0

patmoody:


patmoody:
Oops - wrong branch

PR #389, opened 2017-12-06
PRESIDECMS-1076 Asset manager Lucee 5 - Element at position [1] does …

patmoody:
…not exist in list Array index out of range: 0


DominicWatson:
Thanks for this @patmoody. I'm looking at this and can't see any reason whatsoever for the group by. Am doing some testing but going to switch out for autoGroupBy=true and put this as a fix for 10.8.0. 

Going to close this PR as its targeting 10.9.0 but big thanks for digging into this. Will let you know when fix is available.

PR #390, opened 2017-12-18
PRESIDECMS-1024 Re-send emails

sebduggan:
Also implements PRESIDECMS-1090 - Allow taskmanager tasks to be switchable by feature flag


system/handlers/admin/emailCenter/Logs.cfc
```diff
@@ -50,6 +55,27 @@ component extends="preside.system.base.AdminHandler" {
 		);
 	}
 
+	public string function resendEmail( event, rc, prc ) {
```
DominicWatson: Picky, but we should rename this to `resendEmailAction` + we should ensure that it creates an audit log entry.

DominicWatson:
Awesome sauce, thanks @sebduggan :D

DominicWatson:
This has been merged into upcoming 10.9.0 release.

PR #391, opened 2017-12-20
escaping db columns that are reserved words in ms sql

jjannek:
_user_ and _read_ are both reserved words in MSSQL SQL queries and therefore need to be escaped.


DominicWatson:
Brill, thanks for this @jjannek

PR #392, opened 2017-12-21
PRESIDECMS-1091

john5onCheng:
Corrected the form builder parameter values
Add minlength to textinput and textarea
Corrected matrix input value


DominicWatson:
Brilliant, thanks for these @john5onCheng

PR #393, opened 2018-01-02
PRESIDECMS-1093

john5onCheng:
Apply default group by

### DIFF
--- a/system/services/presideObjects/PresideObjectService.cfc
+++ b/system/services/presideObjects/PresideObjectService.cfc
@@ -218,9 +218,9 @@ component displayName="Preside Object Service" {
 		args.groupBy     = _autoAliasBareProperty( args.objectName, args.groupBy, args.adapter );
 		if ( !Len( Trim( args.groupBy ) ) && args.autoGroupBy ) {
 			args.groupBy = _autoCalculateGroupBy( args.selectFields );
-			if( !Len( Trim( args.groupBy ) ) && !isEmpty( args.extraFilters ) ) {
-				args.groupBy = args.objectName & '.' & getIdField( args.objectName );
-			}
+		}
+		if ( !Len( Trim( args.groupBy ) ) && args.autoGroupBy ) {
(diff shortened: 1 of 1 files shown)


DominicWatson:
Hey @john5onCheng thanks for this.

This logic does not seem correct to me, though. Automatically adding a groupBy just on the id field  whenever there are extra filters does not make sense + will break in non-mysql DBs in many cases.

What is it that is trying to be fixed here? I think an alternative approach is required.

john5onCheng:
hi @DominicWatson,

I have made the changes to always return the groupBy with object id when autoGroupBy is true and autoGeneratedGroupBy value is empty

Many thanks
Johnson

DominicWatson:
But this is just repeating and over-writing the logic in the line before... (i.e. see the function `_autoCalculateGroupBy()`)

It sounds like the logic in that function is what will need updating so that it takes filters as well as select fields into account.

john5onCheng:
hi @DominicWatson ,

The case i am working on is to group by the "website user" when it has query on "contact product" (many-to-many) table through the "contact" object.

Therefore, taking the contact product does not able to provide direct relationship to form a group by for website_user.id. I am not sure is that correct to always return group by value as long as it's not SQL aggregate function when "autogroupby" is set to true when performing selectData()?

Many thanks
J

DominicWatson:
Thanks for this @john5onCheng. This still makes me uneasy though. Doing this will change many existing queries, potentially unpredictably. If we want to force a query to group by everything then we should make this a separate argument:

```
selectData(
    // ...
    , forceAutoGroupBy = true
);
```

Only when this is set to true (default it to false), should we _always_ group by all fields other than aggregates. Does this make sense?

DominicWatson:
Reviewed this more closely. The issue is in fact resolved with a simple `distinct=true` on the relevant selectData() calls (any selectData() call that uses dynamic filters).

DominicWatson:
Thank you for the time looking into it though @john5onCheng

PR #394, opened 2018-01-05
PRESIDE-1098

john5onCheng:
Corrected the default returned value when session is not enabled


DominicWatson:
Great, thanks @john5onCheng. Released in 10.8.62

PR #396, opened 2018-01-08
PRESIDECMS-1102

john5onCheng:
Corrected min length value


DominicWatson:
Yikes, nice one - thanks @john5onCheng!

PR #397, opened 2018-01-08
fixed one time scheduled email template sending

jjannek:
query to find relevant templates did not look after null values for flag indicating that schedule was sent


DominicWatson:
Brilliant, thanks for this @jjannek

DominicWatson:
Fixed in 10.8.63 release. See: https://presidecms.atlassian.net/browse/PRESIDECMS-1104

PR #399, opened 2018-01-09
translates the dynamic filter expressions into German

nueffl:
this PR adds a translation into German for the dynamic filter expressions


sebduggan:
Thanks @nueffl !

PR #400, opened 2018-01-11
PRESIDECMS-1110: MSSQL does not work with MAX function on a bit column type

philippPSH:
My first PR and hope it works. 
I tried the perform the unit tests but had technical problems with it (out of memory 2GB RAM)  and must investigate (spend more time).

I used an MSSQL 2016 database server and the problem came up while login with a new backend user.
Before it was used withe sysadmin only.

### DIFF
--- a/support/tests/integration/api/security/PermissionServiceTest.cfc
+++ b/support/tests/integration/api/security/PermissionServiceTest.cfc
@@ -277,9 +277,9 @@ component output="false" extends="tests.resources.HelperObjects.PresideTestCase"
 		permsService.$( "listUserGroups" ).$args( user="me" ).$results( [ "somegroup", "anothergroup" ] );
 
 		mockContextPermDao.$( "selectData" ).$args(
-			  selectFields = [ "Max( granted ) as granted", "context_key" ]
+			  selectFields = [ "granted", "context_key" ]
 			, filter       = { context = "someContext", permission_key = "a.new.key", security_group = [ "somegroup", "anothergroup" ] }
-			, groupBy      = "context_key"
+			, orderBy      = "context_key, granted"
 			, useCache     = false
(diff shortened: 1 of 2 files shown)


DominicWatson:
This looks good @philippPSH - nice one. Can I double-check your reasoning here as it is a tough one to test? 

You may now end up with results like:

```
| granted | context_key |
=========================
| 0       | keya        |
-------------------------
| 0       | keya        |
-------------------------
| 1       | keya        |
-------------------------
| 0       | keyb        |
-------------------------
| 0       | keyb        |
-------------------------
```

Looping through and setting the struct as we do per context key means we still end up with the same result because you have ordered by `granted` _asc_.

Was that your thinking?

philippPSH:
Hi,
the intention is to have a sorting by "context_key" and "granted"
The resulting query would list the granted=1 entry at the bottom.
(before, MAX (granted) was used to select granted=1 in case of both cases exists).

granted | context_key
-------- | ------------
 0          | keya      
 0          | keya       
 0          | keya        
 1          | keyb        


Do other projects have plenty entries in this table? We do not have an entry in the table  "psys_security_context_permission". We are using the backend only without any permission settings currently.

DominicWatson:
Yes, some projects use this. For example, limit permissions to an object in datamanager. But what you say is exactly right 👍 will merge soon. Thanks for this.

jjannek:
maybe to remove the duplicates it makes sense to add distinct = true?

DominicWatson:
Great stuff, thanks @philippPSH. Merging into 10.8.66 (will also go into latest 10.9). @jjannek I'm hesistant with the distinct as that can cause more performance problems than it solves sometimes. I suggest we leave as is and revisit if anyone should have complaints (the numbers of records are very most likely to be very small and the calculation is cached).

PR #401, opened 2018-01-19
PRESIDECMS-1119 Aligned attribute value to the used value in the adapters

philippPSH:
changed  attributes 
from: ondelete="set-null-if-no-cascade-check" 
to: ondelete="set-null-if-no-cycle-check"

value is used in the BaseAdapter and MSSqlAdater to handle MSSQL specific cascading issues


DominicWatson:
Brill thanks for this @philippPSH

PR #404, opened 2018-01-25
PRESIDECMS-1024 Fix query error when resending disabled

sebduggan:
Only query content fields if feature is enabled.


DominicWatson:
👍

PR #405, opened 2018-01-29
PRESIDECMS-1127 Fix date comparison JS validators

sebduggan:


DominicWatson:
Nice one, thanks @sebduggan - releasing in 10.8.70

PR #410, opened 2018-02-04
PRESIDECMS-1135 Hidden grid fields

sebduggan:


system/services/rendering/ContentRendererService.cfc
```diff
@@ -75,6 +79,7 @@ component {
 			rendered = this.render(
 				  renderer = renderer
 				, data     = arguments.data
+				, record   = arguments.record
```
DominicWatson: I think here, we should **not** add a new `record` arg to the `render()` function (but ok here in the `renderField` function). Instead, make use of the `args` struct:

```
rendered = this.render(
      renderer = renderer
    , data     = arguments.data
    , args     = { objectName=arguments.object, recordId=arguments.recordId, record=arguments.record }
);
```

`render()` is more general purpose and already accepts arbitrary `args` and I think better to keep it that way (we also then can change less code here :) )

DominicWatson (changes requested):
Looking good, just one tweak to the approach I think would be good.

DominicWatson:
Nice one, thanks @sebduggan - merging into upcoming 10.9.0 release.

PR #414, opened 2018-02-17
PRESIDECMS-999 Link picker direct link to assets

sebduggan:


DominicWatson:
Nice one, will make some happy people :). Thanks @sebduggan

PR #415, opened 2018-02-19
PRESIDECMS-1147 DM sort fails if label is formula

sebduggan:


DominicWatson:
Nice one - will make some happy people :) Thanks @sebduggan

DominicWatson:
awesome sauce, thanks @sebduggan

PR #417, opened 2018-02-22
PRESIDECMS-1154 Exclude formula fields from batch edit

sebduggan:


DominicWatson:
Nice one, thanks @sebduggan - will be released in 10.8.78

PR #418, opened 2018-02-23
PRESIDECMS-1157 Do not save file binary to audit log

sebduggan:


DominicWatson:
Fab, thanks @sebduggan. Will be in 10.9

PR #419, opened 2018-02-26
PRESIDECMS-1159 Rebuild/Resend to original address if no recipient ID

sebduggan:


DominicWatson:
Merged, thanks Seb :)

PR #426, opened 2018-03-10
PRESIDECMS-1100 Fixed the Incorrect error message and theme.

cfmitrah:


DominicWatson:
Brilliant, thank you @cfmitrah will be released with 10.8.79.

PR #427, opened 2018-03-14
Fixed PRESIDECMS-879 and PRESIDECMS-1129

cfmitrah:


DominicWatson:
Brilliant, thanks @cfmitrah ! Will be released in 10.8.80 hotfix

PR #428, opened 2018-03-14
removed invalid group detection method calls from mssql 2012 adapter …

jjannek:
…(related to PRESIDECMS-1080)


DominicWatson:
Great, thanks @jjannek

PR #429, opened 2018-03-15
PRESIDECMS-1168

john5onCheng:
Fixed the versioning number not showing up on site tree page


DominicWatson:
Brilliant, thanks @john5onCheng - will be in the next 10.9.0 snapshot.

PR #432, opened 2018-03-19
PRESIDE-1153 - Handle double encoded URL strings for assets - specifi…

nodoherty:
…cally replicable on iOS/iPhone devices


sebduggan:
Great - thanks @nodoherty !

PR #433, opened 2018-03-19
PRESIDECMS-1173: check for non-empty trimmed varchar/text fields befo…

tarekabou-chanab:
…re versioning


DominicWatson:
Nice one, thanks @tarekabou-chanab

PR #435, opened 2018-03-22
Fix edit websiteuser permission

jjannek:
Wrong permissions was evaluated when editing a website user in the Preside admin.


sebduggan:
Thanks @jjannek !

PR #436, opened 2018-03-22
Feature presidecms 1180 form builder no grou java.lang.index out of bounds exception error

seonpks:
fixed the java error due to user insert a value such as '$6' cause java interprets this as a reference to a capturing group, and looks for that group in the regex match


DominicWatson:
Awesome, thanks @seonpks

PR #437, opened 2018-03-23
Fixed the issue that Not possible to de-select a group from the site …

cfmitrah:
…tree permission management


DominicWatson:
Brilliant, thanks for this @cfmitrah ! that was a tricky one :)

PR #438, opened 2018-03-23
feature_PRESIDECMS-1047

cfmitrah:
Added the yesNoSwitch for staying on the edit page, while editing the page from administrator.


system/handlers/admin/SiteTree.cfc
```diff
@@ -264,6 +264,7 @@ component extends="preside.system.base.AdminHandler" {
 		prc.page         = _getPageAndThrowOnMissing( argumentCollection=arguments, allowVersions=true );
 		prc.canPublish   = _checkPermissions( argumentCollection=arguments, key="publish", pageId=pageId, throwOnError=false );
 		prc.canSaveDraft = _checkPermissions( argumentCollection=arguments, key="saveDraft", pageId=pageId, throwOnError=false );
+		rc._backToEdit   = cookie.sitetree_editPage_backToEdit ?: false;
```
DominicWatson: I'd suggest the following here is more robust:

```
rc._backToEdit = IsTrue( cookie.sitetree_editPage_backToEdit ?: "" );
```

`IsTrue()` is a preside helper function for evaluating truthy values.

DominicWatson: Also, please use `cookieService` to get / set cookie variables:

```
rc._backToEdit = IsTrue( cookieService.getVar( "sitetree_editPage_backToEdit", "" ) );

// ...

cookieService.setVar( name="sitetree_editPage_backToEdit", value=true );
```

system/handlers/admin/SiteTree.cfc
```diff
@@ -382,11 +383,16 @@ component extends="preside.system.base.AdminHandler" {
 		);
 
 		getPlugin( "MessageBox" ).info( translateResource( uri="cms:sitetree.pageEdited.confirmation" ) );
-
-		if ( _isManagedPage( page.parent_page, page.page_type ) ) {
-			setNextEvent( url=event.buildAdminLink( linkto="sitetree.managedChildren", querystring="parent=#page.parent_page#&pageType=#page.page_type#" ) );
+		cookie.sitetree_editPage_backToEdit = false;
+		if ( Val( event.getValue( name="_backToEdit", defaultValue=0 ) ) ) {
```
DominicWatson: We now prefer just using `rc` scope rather than `event.getValue()`. So:

```
if ( IsTrue( rc._backToEdit ?: "" ) ) {
// ...
}
```

(There is some older code that uses event.getValue(), but the elvis operator makes it cleaner to just use `rc`).

system/i18n/cms.properties
```diff
@@ -302,6 +302,7 @@ sitetree.add.another=Add another
 sitetree.pageAdded.confirmation=Page added successfully
 sitetree.editPage.title=Edit page, '{1}'
 sitetree.editPage.crumb=Edit page
+sitetree.editPage.backToEdit=I want to edit the page, after saving.
```
DominicWatson: Perhaps: "Keep me on this page" is cleaner?

DominicWatson (changes requested):
Thanks for this @cfmitrah! Logic looking great, have added some review comments for tidying some things up.

cfmitrah:
@DominicWatson Made the changes as per you comments above :)

PR #443, opened 2018-03-28
Feature presidecms 1132

cfmitrah:


cfmitrah:
@DominicWatson @sebduggan Did you get some time to review this feature PR?

sebduggan:
Not had time to look at it yet, but will do soon...

cfmitrah:
Sooner is better to avoid merge conflicts :)

sebduggan (changes requested):
The file views/admin/datamanager/groupVersionNavigator.cfm seems to be missing...

cfmitrah:
@sebduggan  We've added the missing file views/admin/datamanager/groupVersionNavigator.cfm
Please check that.

Thank you

system/handlers/admin/DataManager.cfc
```diff
@@ -1035,6 +1035,53 @@ component extends="preside.system.base.AdminHandler" {
 		return renderView( view="admin/datamanager/versionNavigator", args=args );
 	}
 
+	private string function groupVersionNavigator( event, rc, prc, args={} ) {
```
DominicWatson: I do not think that this belongs in DataManager.cfc. It seems completely specific to system config so should go int he sysConfig admin handler I think.

system/handlers/admin/DataManager.cfc
```diff
@@ -1035,6 +1035,53 @@ component extends="preside.system.base.AdminHandler" {
...
+		var selectedVersion = Val( args.version ?: "" );
+		var objectName      = args.object ?: "";
+		var id              = args.id     ?: "";
+		var setting         = structKeyList( event.getCollectionForForm( "system-config.#args.id#" ) );
+		var maxVersion      = presideObjectService.getNextVersionNumber();
```
DominicWatson: This should be avoided. This will grow the version table unnecessarily. Instead of this (and the query of queries below), we should add a method to the sysConfig service that calculates the max version for a category. This should be achievable with a single regular selectData query I should think. Rough pseudocode:

```
var version = $getPresideObject( versionObjectNameForSysConfgTable ).selectData(
      selectFields = [ "Max( _version_number ) as max_version_number" ]
    , filter = { category = arguments.category }
);

return Val( version.max_version_number );
```

system/handlers/admin/SysConfig.cfc
```diff
@@ -149,6 +160,42 @@ component extends="preside.system.base.AdminHandler" {
...
+			prc.category = systemConfigurationService.getConfigCategory( id = categoryId );
+		} catch( "SystemConfigurationService.category.notFound" e ) {
+			event.notFound();
+		}
+	}
+
+	public void function getConfigHistoryForAjaxDataTables( event, rc, prc ) {
+		prc.setting = structKeyList( event.getCollectionForForm( "system-config.#rc.id#" ) );
+
+		var nextVersionNumber = presideObjectService.getNextVersionNumber();
```
DominicWatson: See other review comment about avoiding this. This method call inserts a new version number in the version number table. Instead, we shoud calculate the real max version number for the settings category.

system/services/presideObjects/PresideObjectService.cfc
```diff
@@ -1260,6 +1261,10 @@ component displayName="Preside Object Service" {
 			args.filterParams = { "#idField#" = arguments.id, _version_changed_fields = "%,#args.fieldName#,%" };
 			args.delete( "fieldName" );
 			args.delete( "id" );
+		} else if ( listLen( arguments.id ) GT 1 ){
```
DominicWatson: Can we remove this? This is an unrelated change. You can achieve the same thing with:

```
selectData( filter={ id=arrayOfIds } );
```

This could be a useful approach, but I'd rather make that a separate change and rather use an array than a string list to avoid ambiguity.

system/views/admin/datamanager/groupVersionNavigator.cfm
```diff
@@ -0,0 +1,62 @@
+<cfscript>
```
DominicWatson: Should be renamed and moved to sys config views folder (see previous comment on the DataManager.cfc)

DominicWatson (changes requested):
Hey guys, this looks great (have tried it out locally). I have a few review comments though about the code.

Would be great to go through those and ammend before we merge.

Great effort though, thank you!

sebduggan:
One other small code style point: when writing cfscript, it's preferable to use <, >, ==, != etc, rather than LT, GT, EQ, NEQ...

system/services/presideObjects/PresideObjectService.cfc
cfmitrah: @DominicWatson 
This is the line called and send configIds,
\system\handlers\admin\SysConfig.cfc : 193, Here if I implement the above changes means I've to change the dataManager and dataManagerService.cfc. Or else we've create a new service which we've to achieve this changes.
BTW Not sure whether we need to implement this change.

DominicWatson:
Hey @cfmitrah - thanks for the changes. I'm going to make a call to leave this out of 10.9. When testing this there were quite a few things not quite working right and as its complex I'd like us to spend the time to get it right before rushing into release.

Will provide more feedback after release.

DominicWatson:

PR #445, opened 2018-04-06
PRESIDECMS-1197 Includes validation of array element 1 content

vivienwong:
Validate array content - element 1 could contain empty string


sebduggan:
Thanks @vivienwong !

PR #446, opened 2018-04-06
PRESIDECMS-1198 Add handler to flag result as false when validation f…

vivienwong:
PRESIDECMS-1198 Add handler to flag result as false when validation failed so that problematic queued record can be removed


sebduggan:
Thanks @vivienwong !

PR #451, opened 2018-04-09
PRESIDECMS-1202: get records before getting record counts for pagination

tarekabou-chanab:
The issue here was `_getPresideObjectService().getRecordVersions()` just returned:
`return selectData( argumentCollection = args );` but because we were using `recordCountOnly=true`, that meant a numeric was being returned instead.

`getRecordVersions` is being used elsewhere so didn't want to change it


sebduggan:
Closing, as issue is already solved by a change in 10.9, which I'll back-port to 10.8.

PR #453, opened 2018-04-12
PRESIDECMS-1205

john5onCheng:
Correction on function name to avoid page lead to 404


sebduggan:
Thanks @john5onCheng - issue was introduced by a merge conflict in the previous release...

PR #460, opened 2018-04-17
PRESIDECMS-898 : Avoid displaying sort children for page with no sub …

cfmitrah:
…pages


DominicWatson:
Awesome sauce, thanks @cfmitrah ! Will merge into upcoming 10.9

PR #466, opened 2018-04-23
PRESIDECMS-1213 : Fixed the side navBar text issue.

cfmitrah:


DominicWatson:
Hey @cfmitrah thanks for this :). What happens in the case of applying this change? Does the nav item then work on two lines? Are you able to upload a screen shot here?

Thanks again, really appreciated.

cfmitrah:
Hi @DominicWatson  @sebduggan 
BTW **before our changes** the nav bar looks like this when we've select French language
beforethechanges

After the changes applied the nav bar displays the text without breaking
afterthechanges


Thanks

sebduggan:
Thanks for this @cfmitrah - I took what you did and ran with it a bit so the text indentation lines up on subsequent lines...

PR #468, opened 2018-04-25
PRESIDECMS-1218 : Fixed the Missing child count key

cfmitrah:


sebduggan:
Thanks for the quick fix @cfmitrah !

PR #470, opened 2018-04-25
PRESIDECMS-990 : Added feature to display the custom message in datat…

cfmitrah:
…able when there is no records to display.


sebduggan:
Great work, @cfmitrah! Merged into 10.9.0...

PR #474, opened 2018-04-27
PRESIDECMS-749 : Fixed the grid fields issue having extra spaces.

cfmitrah:


sebduggan (changes requested):
This is quite a verbose way of doing this. The same can be achieved simply with:

return listToArray( fields, ", " );

This sees both commas and spaces as separators...

cfmitrah:
@sebduggan  I guess that the space will comes for each grid fields like ('birth year') So only i added like that, Hope the DB columns has not having spaces. Your guess is right. I change that.

Thank you

sebduggan (approved):

sebduggan:
Thanks @cfmitrah!

I think if we have spaces in our db column names, then we're going to have much bigger problems than this...

PR #477, opened 2018-04-30
Feature presidecms 1223 dirty form check for ckeditor

nelsonchuah:
Hi Dom,

Have added CKEditor's diryForm check.
Could you help check if i have added correctly, tested and works correctly

Thank you
Nelson


sebduggan:
Hi @nelsonchuah 

We're already adding an onunload listener which calls the protectionListener function, so we should probably use that rather than add another one...

I'd suggest that we create a new function which checks for dirty ckEditor instances, and then call something like:

if ( $form.data( "_isDirty" ) || dirtyRichEditors() ) {

sebduggan:
Thanks @nelsonchuah - merged into 10.9.0 with a small modification

PR #478, opened 2018-04-30
PRESIDECMS-1231 Add useCache argument to AjaxSelect

jbez:
Recently stumbled across this. Would be nice to be able to bypass query caching for AjaxSelects this way.


DominicWatson:
This would be great, yes, thanks for the PR :). I think it would also make sense as a configurable option on the object picker, so that you could do the following in your form definitions:

```
<field ... useCache="false" />
```

This option would then feed through to the ajax URLs for that form control instance, passing through to `datamanager.getObjectRecordsForAjaxSelectControl` handler that could then set this when calling the service method that you have changed.

Do you think this is something you could take a look at while you're here? The form control that would require changing can be found at `/handlers/formcontrols/ObjectPicker.cfc`.

jbez:
This should do the trick. 

useCache defaults to 'true' for ObjectPicker form-fields and their respective ´getObjectRecordsForAjaxSelectControl´ and can be bypassed using ´useCache=false´ inside the form definitions.

system/handlers/admin/DataManager.cfc
```diff
@@ -198,6 +198,7 @@
 			var filterValue    = "";
 			var orderBy        = rc.orderBy       ?: "label";
 			var labelRenderer  = rc.labelRenderer ?: "";
+			var useCache	   = structKeyExists(rc, 'useCache') ? rc.useCache : true;
```
DominicWatson, line 201: FYI: Tweaking this to the more concise:

```
var useCache = IsTrue( rc.useCache ?: "" );
```

`IsTrue()` is a coldbox helper function defined in Preside core so available to handlers and views. It ensures that the result is boolean, no matter the user input.

DominicWatson:
Super, thanks @jbez :)

FYI, I tidied up some whitespace (we use spaces for alignment, tabs for indentation - so no tabs after the first non-whitespace character in a line) and also a little trick to get boolean values from user input: `IsTrue( rc.somevariable ?: "" )`.

Also, you might want to checkout the contrib guide with regards to Git branching: https://beta-docs.preside.org/contribguides/submittingchanges.html. If you do pull requests from your stable branch you could end up in a mess when wanting to do further PRs, etc. Helps to have a separate branch for each PR.

But anyways, really appreciate the contrib, will make it into the 10.9 release :)

DominicWatson:
Manually closing this as already merged - something gone wrong with the PR process.

PR #484, opened 2018-05-03
PRESIDECMS-1230

john5onCheng:
Add parameter name so that it always return with given default value instead of empty string from session storage getVar function

Hi @DominicWatson,

Could you help to merge into 10.8 and 10.9?

Many thanks
J


DominicWatson:
Super, thanks @john5onCheng :)

DominicWatson:
Merged into its own hotfix: 10.8.85

PR #486, opened 2018-05-04
Feature presidecms 1219 timeout when editing assset manager folder permissions

nelsonchuah:
Hi Dom,

Have moved editfolder action to run in adhoc task

Thank you
Nelson


system/handlers/admin/AssetManager.cfc
```diff
@@ -340,13 +340,17 @@ component extends="preside.system.base.AdminHandler" {
...
 		}
 
-		try {
-			assetManagerService.editFolder( id=folderId, data=formData );
-		} catch ( any e ) {
-			logError( e );
-			messageBox.error( translateResource( "cms:assetmanager.edit.folder.unexpected.error" ) );
-			setNextEvent( url=event.buildAdminLink( linkTo="assetManager.editFolder", querystring="folder=#parentFolder#&id=#folderId#" ), persistStruct=formData );
-		}
+		var taskId = createTask(
```
DominicWatson: I think this is not quite the right place to do this. We only really need to operate in a bg thread if/when we need to move files between public and private storage.

The general act of updating folder metadata should be done in the same thread. The change would be better off inside `AssetManagerService.cfc$editFolder()`:

```
if ( data.keyExists( "access_restriction" ) && folder.access_restriction != arguments.data.access_restriction ) {
	ensureAssetsAreInCorrectLocation( folderId=arguments.id ); // this line should be changed so that an ad-hoc task is created here
}
```

The adhoc task handler action should then run `assetManagerService.ensureAssetsAreInCorrectLocation( ... )`.

Make sense?

DominicWatson (changes requested):
Hi Nelson, thanks for this. Please see review feedback. Wants tweaking a little I think.

nelsonchuah:
Hi Dom,

Yup that make more sense to run only when folder's permission is changed.

Have updated the code accordingly

Thank you
Nelson

DominicWatson:
Awesome sauce, thanks @nelsonchuah :)

PR #487, opened 2018-05-07
Feature presidecms 884 resend token

annbelievable:


system/handlers/Tasks.cfc
```diff
@@ -18,4 +19,17 @@ component {
 	private boolean function processBatchedEmails( logger ) {
 		return emailMassSendingService.processQueue( arguments.logger ?: NullValue() );
 	}
+
```
DominicWatson: I'm not sure that we should send this out on a schedule. I think more preferable would be to resend only when they attempt to reset their password with an expired token.

DominicWatson (changes requested):
Hey @annbelievable, thanks for this, looking good :)

A functional change I'd like to make, however, is that I don't think should be a scheduled task. Instead, it should trigger when a user attempts to reset password with an expired token.

The message would then say something like: "Your token has expired, we've sent you another one..." (you wouldn't need to change anything here, this would be editorial, of course).

annbelievable:
Removed the scheduled task, send out the email once an expired token is detected (when validating the reset password token).

DominicWatson:
Nice one, thanks @annbelievable. I've created a new feature branch in Preside (`feature-PRESIDECMS-884_resend_token`) and made a few changes (renamed some things, fixed some tests, etc.).

It would be great to start a new Pull request from and to that new branch with following ammendments:

* Add the same functionality to the admin
* Allow turning the feature on/off using a feature flag. Default should be "off". This is because any existing messaging around token expiry will be aligned with not using this feature so those upgrading will have a "broken" user experience

PR #488, opened 2018-05-08
Feature presidecms 1234 notification performance improvement

nelsonchuah:
Hi Dom,

Have done a few updates to Preside's notification

1) personal preference tabs are now on the left
2) by default preside users won't be subscribed to all notification
3) notification setting can be navigated from system setting dropdown menu now
4) schedule task to delete old notifications (record deleted from admin_notification & admin_notification_consumer table) by default notification will be kept for 12 months (can be edited from system setting)

Thank you
Nelson


system/services/notifications/NotificationService.cfc
```diff
@@ -527,6 +527,37 @@ component autodoc=true displayName="Notification Service" {
 		);
 	}
 
+	public boolean function deleteOldNotifications( any logger ) {
+
+		var keepNotificationsFor = Val( $getPresideSetting( "notification", "keep_notifications_for_months", 12 ) );
```
DominicWatson: It would be good to either have an additional "On/Off" switch, or allow a value of `zero` for this setting to be able to never delete old notifications.

system/services/notifications/NotificationService.cfc
```diff
@@ -527,6 +527,37 @@ component autodoc=true displayName="Notification Service" {
...
+		var totalOldNotification = oldNotifications.recordCount()
+
+		if ( !totalOldNotification ) {
+			if ( canInfo ) { logger.info( "No notification to delete." ); }
+			return true;
+		} else {
+
+			if ( canInfo ) { logger.info( "Deleting old notifications..." ); }
+
+			var notificationIds = ValueArray( oldNotifications.id );
```
DominicWatson: Rather than selecting the data first (which could be *very* expensive for first time running) - we should do:

```var notificationsDeleted = $getPresideObject( "admin_notification" ).deleteData( filter="datecreated < :datecreated", ... );```

I _believe_ that the notification consumer records should get cascade deleted automatically so no need for a separate delete on that.

File: system/forms/system-config/notification.xml [hunk 3]
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--#!autodoc
+System config form: notification settings
+
+This form is used for managing settings for auto removal of old notifications
+-->
+<form i18nBaseUri="system-config.notification:">
+	<tab id="default">
+		<fieldset id="default" sortorder="10">
+			<field name="keep_notifications_for_months"   control="spinner"      default="12" />
```
DominicWatson: I think a setting in days makes more sense, here. Allows more granularity and users should still be able to figure out 365 days, or 30 days for a month, etc.

DominicWatson (changes requested):
Hey @nelsonchuah, looking good :). See code review for a few tweaks.

nelsonchuah:
Hi Dom,

Have done the updates.
- have also set it default to be turned off to prevent user's notification to be deleted without them knowing

Thank you
Nelson

DominicWatson:
Fantastic, thanks @nelsonchuah

PR #489, opened 2018-05-08
PRESIDECMS-1194

john5onCheng:
Redirect bot access from admin page to 401


system/base/AdminHandler.cfc
```diff
@@ -8,6 +8,9 @@
 		<cfargument name="eventArguments" type="struct" required="true" />
 
 		<cfscript>
+			if( isBot() ){
```
DominicWatson: The key here really is to check whether or not the request is a "stateless" request, or not. That is controlled in your Application.cfc with:

```super.setupApplication(
    // ...
    , statelessUrlPatterns = [ ... ]
    , statelessUserAgentPatterns = [ ... ]
);```

(or by implementing your own `isStatelessRequest()` method in your Application.cfc). 

You are then able to check whether or not the request is stateless with `event.isStatelessRequest()` so this should be used instead of `isBot()`.

DominicWatson (commented):
Hey @john5onCheng, thanks for this, appreciated. See review feedback - there is some core functionality that should be used instead of a new function :)

john5onCheng:
Updated the PR :)

DominicWatson:
AWesome sauce, thanks @john5onCheng

DominicWatson:
Merging into 10.9.0 - snapshot117

PR #490, opened 2018-05-09
Feature presidecms 805

cfmitrah:


DominicWatson:
I think this could be cool but have doubts about this in core / this implementation.

A couple of things:

* favicons are a little more complex these days than just uploading a single icon against a site
* should not be called `main_image`
* should probably have a dedicated UI

It might be good to start this as a separate extension that adds some more powerful features around favicons (perhaps hooking into APIs that will generate all the various favicons for you based on an uploaded large icon). Other considerations include changing the physical filename of favicons when you make changes to the favicon (to help with browser favicon caching).

I'm closing this for now and will comment on the ticket. Really appreciate the contribution, just not sure this is ready yet.

PR #491, opened 2018-05-11
PRESIDECMS-884 enable admin to have resend token feature and also to …

annbelievable:
enable admin to have resend token feature and also make the feature switchable


DominicWatson:
Brillo - merged into 10.9 :). Thanks @annbelievable

PR #492, opened 2018-05-12
PRESIDECMS-1239 : Added image thumbnail in asset browser

cfmitrah:
I've added image thumbnail in asset browser while picking from CKEditor
asset_browser

### DIFF
--- a/system/views/renderers/asset/icon.cfm
+++ b/system/views/renderers/asset/icon.cfm
@@ -1,6 +1,6 @@
 <cfscript>
 	ext     = ListLast( args.storage_path ?: "", "." );
-	iconUrl = event.buildLink( assetId=args.id ?: "", derivative='pickerIcon' );
+	iconUrl = event.buildLink( systemStaticAsset = "/images/asset-type-icons/16px/#LCase( ext )#.png" );
 </cfscript>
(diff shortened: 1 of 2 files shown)


system/views/renderers/asset/icon.cfm
```diff
@@ -1,6 +1,6 @@
 <cfscript>
 	ext     = ListLast( args.storage_path ?: "", "." );
-	iconUrl = event.buildLink( systemStaticAsset = "/images/asset-type-icons/16px/#LCase( ext )#.png" );
+	iconUrl = event.buildLink( assetId=args.id ?: "", derivative='pickerIcon' );
```
DominicWatson: This isn't quite right, unfortunately. This will break a lot of other places that can't render an image for assets such as PDFs, etc. 

I think the change that needs to happen here is the code that renders the asset for the picker or, the code that you have here should be made in `/renderers/image/icon.cfm` (note the different folder).

DominicWatson (commented):
Thanks for this @cfmitrah. Its not quite right unfortunately. Could you check the review?

DominicWatson:
Awesome sauce guys, thank you!

PR #493, opened 2018-05-14
PRESIDECMS-1235 clearing the cached item type config resolves the issue

annbelievable:
Resetting the variable cachedItemTypeConfiguration to {} inside FormBuilderItemTypesService.cfc is able to resolve the issue.


DominicWatson:
Awesome detective work, thanks @annbelievable :). Will be released with 10.9.0 (just released in 10.9.0-snapshot122).

PR #495, opened 2018-05-16
PRESIDE-1208 Exclude mailto links from being replaced

vivienwong:
Email tracker: Exclude mailto links from being replaced


DominicWatson:
Perfect, thanks @vivienwong

PR #496, opened 2018-05-18
PRESIDECMS-1243

john5onCheng:
Add default sort order for exported data


DominicWatson:
Nice one, thanks @john5onCheng. merging into upcoming 10.9.0 release

PR #497, opened 2018-05-21
Feature presidecms 609 fix default value method

nelsonchuah:
Hi Dom,

Have added a fix for Dynamic default value, to be run before form is render.

A small note when upgrading preside to have the fix below:
- we would need to add default value for variable in object's method, i.e.
```
public string function createSlug( required struct product ) {
        var productLabel = arguments.product.label ?: "";
        ...
        ...
}
```

Thank you
Nelson


system/services/forms/FormsService.cfc
```diff
@@ -1475,6 +1475,29 @@ component displayName="Forms service" {
...
+				break;
+				case "method":
+					var obj = _getPresideObjectService().getObject( arguments.objectName );
+
+					defaultValue = obj[ ListRest( defaultValue, ":" ) ]( {} );
+				break;
+			}
+		}
+
+		return defaultValue;
```
DominicWatson, line 1498: It could make sense here to do:

```
return defaultValue ?: "";
```

This way, if any existing implementations return `null` in some circumstances, they won't error and instead just have no default value which would be expected.

I'll make this adjustement while merging, but thought be good to mention here.

DominicWatson:
Nice one @nelsonchuah - thank you. Merging now.

PR #498, opened 2018-05-21
PRESIDECMS-843 Trigger filtering of related objects after QuickAdd

sebduggan:


DominicWatson:
awesome sauce, thanks @sebduggan

PR #499, opened 2018-05-21
feature_PRESIDECMS-476 ::

cfmitrah:
- Added functionality to convert .tif / .tiff files into .jpg format default.
- Added this feature in both AssetManager and also AssetPicker.


DominicWatson:
Thanks for this guys. We're currently preparing for a 10.9 release and would like to hold off new features until after then. This means it might be a little while until we get to this properly.

Quick feedback though, we should consider how a developer could allow tiffs to be uploaded without converting them to jpg. i.e. perhaps they need to be uploaded raw for some T-Shirt printing company (or similar uses)!

DominicWatson:

PR #500, opened 2018-05-21
PRESIDECMS-850 Reinstate deprecated column from object properties ins…

vivienwong:
Reinstate deprecated column from object properties, not from DB field


DominicWatson:
Brilliant, thank you @vivienwong ! Will be released with 10.9

PR #501, opened 2018-05-22
PRESIDECMS-1247 Remove redundant output=false from components

cfmitrah:
- Cleanup scaffolded code


DominicWatson:
Love this :). Its not quite what PRESIDECMS-499 is about though so I've added a separate task and renamed this pull request (see https://presidecms.atlassian.net/browse/PRESIDECMS-1247). Merging into 10.9.0.

Thank you!

PR #502, opened 2018-05-22
PRESIDECMS-615: background position when admin toolbar showing

sebduggan:


DominicWatson:
Nice one, thanks @sebduggan

PR #506, opened 2018-05-28
feature_PRESIDECMS-699::

cfmitrah:
- Display Active / Inactive status on pages while editing.


DominicWatson:
Thanks for this @cfmitrah . Going to hold off looking at this until after 10.9 release as it introduces a UI change and we're currently prepping the release.

Will pick it up after.

DominicWatson:

PR #507, opened 2018-05-28
Changed Javaloader dependency to use a fix version

jjannek:
The cbjavaloader dependency used to be a fuzzy version (give me any stable version newer than...) which will be problematic in terms of later rebuilding the same Preside package version. If the javaloader package changes in between the resulting Preside package could differ. Therefore I suggest using a fixed version, which is 1.5.0+35 as of now.


DominicWatson:
Nice one, thanks Jan.

PR #509, opened 2018-05-28
feature_PRESIDECMS-915 ::

cfmitrah:
- Remove unwanted \n \t from json ajaxproxy results.

### DIFF
--- a/system/services/helpers/JQueryDatatablesHelpers.cfc
+++ b/system/services/helpers/JQueryDatatablesHelpers.cfc
@@ -24,13 +24,11 @@ component {
 		for( row in arguments.qry ){
 			record = {};
 			for( col in arguments.columns ){
-				record[ ListLast( col, "." ) ] = row[ ListLast( col, "." ) ];
+				record[ ListLast( col, "." ) ] = reReplace(row[ ListLast( col, "." ) ],"\s","","ALL");
 			}
+
 			ArrayAppend( result.aaData, record );
 		}
-
-		if( isStruct( result.aaData[1] ) AND result.aaData[1]._options NEQ '' )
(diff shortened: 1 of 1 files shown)


system/services/helpers/JQueryDatatablesHelpers.cfc
```diff
@@ -29,6 +29,9 @@ component {
 			ArrayAppend( result.aaData, record );
 		}
 
+		if( isStruct( result.aaData[1] ) AND result.aaData[1]._options NEQ '' )
```
DominicWatson: I'm not sure about this. Firstly:

* This presumably should be run against each row
* The cleanup needs to be a little smarter than just removing all whitespace?

I imagine the whitespace cleanup would be something like:

```reReplace( result.aaData[ i ]._options, "\s+", " ", "all" )

(i.e. tidyup double spaces, not get rid of all).

Would be good to see some before and after output after making this fix.

DominicWatson (changes requested):
Hi guys, thanks for this - could you take a look at the review feedback?

cfmitrah:
- Functionality optimized. ( now it's  run against each row )
- yes that functionality removing all white space.
- I've attached screen samples for before and after changes.
**Before change :** 
before

**After change :**

after

Please check that and let me know if any update needed ?

DominicWatson:
The issue with removing _all_ whitespace is that text gets broken. There can be editorial text in the return (as in your example) and all the spaces are removed from that.

So, yes - needs to not remove _all_ whitespace, but rather convert 1 or more consecutive whitespace characters into a single whitespace character.

DominicWatson:

PR #510, opened 2018-05-30
feature_PRESIDECMS-708 ::

cfmitrah:
- Remove "Edit global site settings" button in sitetree view.


DominicWatson:
Hey guys, thank you for this. I'm not sure about this as a change, however (and too late in the day to squeeze into 10.9.0).

I don't really see that the button here is causing harm - and we have permissioning that can avoid issue of wrong people editing what they shouldn't. Going to close this for now but can revisit as a discussion if wanted.

Apologies for not triaging the ticket sooner.

PR #511, opened 2018-05-30
feature_PRESIDECMS-744 ::

cfmitrah:
- Review and enable more MIME Types for the Asset Manager


DominicWatson:
Hey guys, thanks for this. Could you also ensure that all the new file types also have a corresponding entry in:

```/preside/system/i18n/filetypes.properties```

This means that they will then render nicely in the file type picker, etc.

cfmitrah:
Added corresponding filetypes in  ( /preside/system/i18n/filetypes.properties ) file.

DominicWatson:

PR #512, opened 2018-05-30
feature_PRESIDECMS-455 ::

cfmitrah:
- Switch language in site tree (multilanguage sites)


DominicWatson:
Similar to the removal of the manage site button, I'm not sure about this just yet - but thanks for submitting. 

With the datamanager changes we've made we can begin to make these UIs more customizable so that you can add these types of options through and extension, etc.

Closing for now.

PR #513, opened 2018-05-30
feature_PRESIDECMS-1252 ::

cfmitrah:
- Fixed UI overlaps issue.

### DIFF
--- a/system/views/admin/rulesEngine/index.cfm
+++ b/system/views/admin/rulesEngine/index.cfm
@@ -5,7 +5,7 @@
 
 
 <cfoutput>
-	<div class="top-right-button-group" style="margin: -8px 4px 0px 0px;">
+	<div class="top-right-button-group" >
 		<cfif hasCmsPermission( "rulesEngine.add" )>
 			<br>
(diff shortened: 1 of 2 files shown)


system/views/admin/rulesEngine/index.cfm
```diff
@@ -5,8 +5,9 @@
 
 
 <cfoutput>
-	<div class="top-right-button-group">
+	<div class="top-right-button-group" style="margin: -8px 4px 0px 0px;">
```
DominicWatson: Not sure on this one. I think we need a global fix for this issue with the 'top-right-button-group'. i.e. some kind of `event.addAdminTopRightButton()`, `event.getAdminTopRightButtons()` set of methods. These could then be output at the very end of the request in a CSS layout that doesn't require absolute positioning.

Can you remove this bit from the PR and we can then merge the other change?

DominicWatson (changes requested):
Thanks @cfmitrah - please see review comments.

cfmitrah:
> Can you remove this bit from the PR and we can then merge the other change?

Yes I removed that inline style, now  you can merge the other change.

DominicWatson:

PR #514, opened 2018-06-01
PRESIDECMS-1253

john5onCheng:
Exclude to disable object picker class


DominicWatson:
Nice, thanks for the deep investigation and fix @john5onCheng

PR #516, opened 2018-06-01
feature_PRESIDECMS-827

cfmitrah:
- Added Google Tag Manager (GTM) Integration


DominicWatson:
Hey guys, thanks for this. I have a strong feeling that this should be an extension that follows the same pattern as: https://github.com/pixl8/preside-ext-google-analytics.

Apologies for not triaging the ticket sooner and making that clear. However, general rules:

* If a JIRA ticket hasn't been marked as "Accepted", then its worth asking about it before going ahead
* If the feature is about integrating with a proprietary third party (i.e. Google), then it should probably not belong in core

PR #517, opened 2018-06-01
feature_PRESIDECMS-1255 ::

cfmitrah:
- Fixed label value doesn't  show properly issue in Scheduled task configuration page


DominicWatson:
Awesome sauce, thanks @cfmitrah

PR #519, opened 2018-06-04
Renderlabel tenancy bypass

jbez:
Allow bypassing of tenancy for renderLabel.


DominicWatson:
Hi @jbez thanks for this. However, I'm 99% sure we've already merged this into the upcoming 10.9 branch (can see code in there that does this) - is this a repeat? Closing for now but please ping me if you disagree!

PR #522, opened 2018-06-08
PRESIDECMS-1266 Add view version permission checking to view record and listing buttons

nelsonchuah:
Hi Dom,

Have added permission checking for view version to datamanager

Thank you
Nelson


DominicWatson:
Fab, perfectly executed - thanks @nelsonchuah

PR #528, opened 2018-06-15
PRESIDECMS-1270 fixes a bug that caused modal titles to be lowercased…

nueffl:
… by default which causes a grammatical problem in some languages (e.g. German)


DominicWatson:
Nice one, thanks @nueffl . Merged into upcoming release.

PR #529, opened 2018-06-16
feature_PRESIDECMS-476 ::

cfmitrah:
- Added system settings for tiff image handling.
- If tiff image conversion is enabled then .tif / .tiff files convert into .jpg format. otherwise tiff files uploaded normally.
- Added this feature in both AssetManager and also AssetPicker.


DominicWatson:
Super, thanks @cfmitrah - this has been merged into 10.10 waiting release.

PR #532, opened 2018-06-25
feature_PRESIDECMS-1275 ::

cfmitrah:
- Added form validation for Email center scheduled options .


DominicWatson:
Thanks for this @cfmitrah, appreciate the PR. However, the fix was more involved and has been done since.

PR #533, opened 2018-07-05
PRESIDECMS-1339 Added argument bypassTenants to renderLabel

jbez:
i.e. _articles_ published on _pages_ of different _sites_, where _pages_ have tenant="_sites_", 
the labelRenderer needs to be able to bypass the "sites" tenancy to get the label field for the pages object..


DominicWatson:
Fab, releasing with hotfix 10.9.9. Thanks @jbez

PR #536, opened 2018-07-12
PRESIDECMS-1338 OneToManyConfigurator Arguments: add,edit,removable

jbez:
This allows for specifying the available options in oneToManyConfigurator formcontrols.
i.e. Only allow editing of data, but not adding or removing exisiting entries.


DominicWatson:
Nice one, thanks @jbez - will be in 10.9.9 hotfix release.

PR #537, opened 2018-07-20
Feature presidecms 1299 filename encoding for form builder export

choontat-pixl8:
Ensure `toBase64` and `toString` uses UTF-8 encoding instead of pulling default encoding of the page on which the function is called.


sebduggan:
Thanks @choontat-pixl8 !

PR #540, opened 2018-08-02
Feature presidecms 1307 login pages alert msgs

annbelievable:
I have added the policy message in reset password page and password reset confirmation message in login page. I have also added logic to ensure that there will always be a default message.


DominicWatson:
Nice one, thanks @annbelievable. Will be released in 10.9.6 hotfix.

PR #541, opened 2018-08-06
Fix bug on renderview when cachetimeout is not numeric.

jtcy0123:
Fix bug on renderview when cachetimeout is not numeric causing error in _generateCacheKey which requires numeric value


jtcy0123:
the bug is actually on bleeding edge. closing this pull request

PR #542, opened 2018-08-06
Fix bug on _generateCacheKey that requires numeric cacheTimeout 

jtcy0123:
Fix bug on _generateCacheKey that requires numeric cacheTimeout and numeric cacheLastAccessTimeout which are default to empty string if not provided


DominicWatson:
Nice one, thanks @jtcy0123. merging into the bleeding edge 10.10 release :)

PR #546, opened 2018-08-22
PRESIDECMS-1318 Changed default sorting in case pagination is used in MSSQL2012

jjannek:
In case "distinct" is used the current "select 0" concept used as default sorting throws an error.
If changing to "select 1" this means it will automatically sort by the first column in the select list.


DominicWatson:
Nice one, merged into bleeding edge 10.10.0. Thanks @jjannek

PR #547, opened 2018-08-31
added i18n DE translations for cloning

philippPSH:
changed dynamic expression DE transalation for "PropertyIsNull"


philippPSH:
I will wait for the project in zanata for the i18n translations

PR #548, opened 2018-09-12
Feature preside 1335 uploadassets

annbelievable:


DominicWatson:
Nice one, thanks @annbelievable :) Will be released in hotfix 10.9.9

PR #550, opened 2018-09-13
PRESIDECMS-1364 extension types require a '.' in front

annbelievable:
1) the extension types requires a '.' in front or else it wouldnt work
2) uses '=' sign instead of ':' for all the event.includedata


DominicWatson:
Thanks @annbelievable - this will be available in 10.9.14

PR #554, opened 2018-10-19
PRESIDECMS-1366

john5onCheng:
Add template name


DominicWatson:
The fix for this is a little more involved. There is logic in the code that checks for this being an empty string. I believe the fix is better done in the logic - setting the active template to "default" when it is empty. I will take a look at the code and put in a fix asap.

PR #555, opened 2018-10-22
Feature presidecms 1331 restore pending image

john5onCheng:


john5onCheng:
hi @DominicWatson, 
Could you review this changes?

DominicWatson:
Hi @john5onCheng, thanks for this. I think it still has the same issue with the potential to have parallel operations both trying to generate the derivative, but I like the retries idea. I'll take a look to see if I can do something to guarantee that there will not be two threads both trying to generate the same image.

+1

PR #556, opened 2018-11-02
PRESIDECMS-1372 Adhoc tasks should fail gracefully if task not found

sebduggan:


DominicWatson:
Nice one, thanks @sebduggan

PR #557, opened 2018-11-07
Feature presidecms 1374 cms permissions on child pages of child pages

choontat-pixl8:


DominicWatson:
Fantastic, thanks @choontat-pixl8 ! Will be released in 10.10.12

PR #560, opened 2018-12-04
Feature presidecms 1398 formbuilder excel export submitted by fix

nelsonchuah:
Hi Dom,

For Formbuilder excel export submitted_by is displaying id field instead of label, have added renderLabel to submitted_by field during export

Thank you
Nelson


DominicWatson:
Fab, thanks @nelsonchuah . Released in 10.10.12

PR #561, opened 2018-12-04
Feature presidecms 1399 formbuilder email from

annbelievable:


DominicWatson:
I believe this may have been fixed by other fixes @annbelievable . Could you verify that this is working as expected on the latest 10.10 release (10.10.12)?

annbelievable:
Hi Dom,

Just checked, yup, this is fixed in newest release.

Thank you

PR #563, opened 2018-12-06
PRESIDECMS-1404: cast text to varchar for MSSQL

smoufa:
When MSSQL is used the script now casts the TEXT type to VARCHAR before comparing.


DominicWatson:
Nice one, thanks @smoufa. Will be released in 10.10.13

PR #564, opened 2018-12-11
Presidecms 1408 mssql cast fix

smoufa:
Fixing another occurence of db type text used as varchar without casting. Also fixed the box.json version number so that db migration scripts run when project is started and preside is used without package manager.


DominicWatson:
Nice one, thanks @smoufa !

PR #569, opened 2018-12-28
PRESIDECMS-1419

john5onCheng:
Hide recipient if filterObject is empty


DominicWatson:
Nice one, thanks @john5onCheng . Will be released with 10.10.17 hotfix.

PR #570, opened 2019-01-03
PRESIDECMS-1427 Add missing subtitle for oneToMany.addRecord.page in DataManager

smoufa:
DataManager.addOneToManyRecord tries to resolve a label that doesn't exist causing a `**NOT FOUND**` to be displayed


DominicWatson:
Nice find, thanks @smoufa. Will be released in 10.10.19 (I missed merging into 10.10.18 by mistake - doh!)

PR #571, opened 2019-01-04
PRESIDECMS-1425 Remove obsolete quote

smoufa:
I removed an obsolete quote in the view.

Generated before:
`<a class=" abc123"" data-context-key="s">`

After:
`<a class=" abc123" data-context-key="s">`


DominicWatson:
Nice find. Thanks @smoufa. Will be released in 10.10.18

PR #573, opened 2019-01-10
PRESIDECMS-1424 Read from the latest version for page type fields

vivienwong:
Obtain the correct version when multilingual consists of both page and page_type fields


DominicWatson:
Superb, thanks for the investigation and fix @vivienwong

PR #574, opened 2019-01-14
PRESIDECMS-1431: Fix for data export not auto grouping records

jjannek:
data export should use autoGroupBy and not autoGroup (which does not exist)


DominicWatson:
Fab, nice pickup @jjannek . Note I created JIRA ticket and renamed issue to link it. Will be released with 10.10.19

PR #575, opened 2019-01-16
Feature presidecms 1431 data manager sort by formula field

choontat-pixl8:


DominicWatson:
Fantastic, thanks for this tricky fix @choontat-pixl8 :) Will be released in 10.10.19

PR #576, opened 2019-01-16
PRESIDECMS-1432 Full page caching preserve content type

sebduggan:


DominicWatson:
Awesome sauce, thanks for the fix @sebduggan. Will be released in 10.10.19 hotfix

PR #578, opened 2019-01-18
PRESIDECMS-1436

john5onCheng:
Show folder with hidden = null value to select folder list


DominicWatson:
Nice one, thanks for this @john5onCheng . Will be released in 10.10.21

PR #582, opened 2019-01-24
PRESIDECMS-1448

madmikede:
prefer Java frowar request URI before cgi.path_info


DominicWatson:
Nice one - thanks @madmikede :) :)

PR #584, opened 2019-01-30
PRESIDECMS-1451

john5onCheng:
Allow CMYK image get uploaded with no error when Image magick is turned on


DominicWatson:
Nice one, thanks for this @john5onCheng . Has been around a while - great one to fix!

PR #585, opened 2019-01-30
PRESIDECMS-1452 param for missing request variable

madmikede:
The fix for Apache2/Tomcat constellation may fails if request[ "javax.servlet.forward.request_uri" ] doesn't exist. Added a defaultvalue.


DominicWatson:
Nice one, thanks @madmikede will be in 10.10.26

patmoody:
Well done @madmikede nice find! This may help us.

PR #591, opened 2019-02-14
PRESIDECMS-1473 Set filename as asset title when not submitted in assetData

madmikede:
When adding assets programmatically with addAsset and having two files with the same file name the current code returns an error. The problem is, that asset.title is used in in _ensureUniqueTitle(). If it is missing in the assetData struct it will never be modified to a unique name. The solution is setting the asset.title to arguments.fileName if it not set.


DominicWatson:
Awesome, thanks @madmikede. Will be released with 10.10.37

PR #594, opened 2019-02-18
PRESIDECMS-1476 Impersonation refactoring

sebduggan:


system/services/websiteUsers/WebsiteUserImpersonationService.cfc
```diff
@@ -0,0 +1,74 @@
...
+
+		return impersonation.targetUrl;
+	}
+
+
+// private methods
+	private string function _addImpersonation( required string userId, required string targetUrl ) {
+		var impersonationId = createUUID();
+
+		variables.impersonations[ impersonationId ] = {
```
DominicWatson: The only issue with using variables scope for this type of caching is that this feature will not be clusterable. Some variables caching of features are ok, because they are just warming up computation caching that will be consistent across all nodes - but in this case, you would not be able to have Preside in a cluster and configure this impersonation to work consistently. 

Suggest refactoring so that the service accepts a cachebox cache and uses that to store the impersonations. (if using a cluster, can then configure the cache to use an eternal cache provider).

DominicWatson (changes requested):
Looking good, but for small detail re storage of impersonations.

system/services/websiteUsers/WebsiteUserImpersonationService.cfc
sebduggan: Done and committed...

DominicWatson:
Awesome sauce, merging into 10.10.39

DominicWatson (approved):

PR #595, opened 2019-02-20
PRESIDECMS-1479 Filter on label renderer labels

sebduggan:


DominicWatson:
This is a beautiful thing. Thanks @sebduggan, will be available in 10.10.40

PR #597, opened 2019-02-21
PRESIDECMS-1482 Redirect to page not found if record does not exist

vivienwong:
Deleted record should not be shown in view record


DominicWatson:
Nice one, thanks @vivienwong - will be released with 10.10.41

PR #601, opened 2019-03-01
PRESIDECMS-1488

john5onCheng:
Set default max day to keep email log


DominicWatson:
These values should be configurable in email settings and _default_ to 90 days. I have done that now so good to close.

Can see the commit here: https://github.com/pixl8/Preside-CMS/commit/f0ebba16bf61f3414ca57a759810cb4448d43a61

Thanks for the PR.

PR #604, opened 2019-03-13
PRESIDECMS-1501: Render audit log message for objects without label

smoufa:
This would now show the recordId as the link label if no label can be generated.


DominicWatson:
Awesome sauce, thanks @smoufa - merging into 10.10.49

PR #609, opened 2019-03-27
PRESIDECMS-1527 update ckeditor

sebduggan:


DominicWatson:
Does the update have any particular positive effects? (realise that it is good to keep up to date, but does this address paste from word issues, for example, other than the margin change)?

sebduggan:
There are a number of Paste From Word fixes in the release notes since the previous version we were using, and it seems to work a bit better now...

The changes since then have mostly been bug fixes, so I think it's probably a good thing.

PR #616, opened 2019-04-03
Presidecms 1535 extra filter sort

vivienwong:
Sort first level tree records with custom filter - for Preside BE (10.11)


DominicWatson:
Nice one, thanks @vivienwong, have merged into upcoming 10.11.0 release.

PR #621, opened 2019-04-15
PRESIDECMS-1547 More helpful collation error message

sebduggan:


DominicWatson:
Nice one, thanks Seb. Merged into 10.10.62

PR #622, opened 2019-05-03
PRESIDECMS-1553 Error handling on adhoc task http call

sebduggan:


DominicWatson:
Nice one, thanks Seb. Released in 10.10.62

PR #627, opened 2019-05-15
Feature presidecms 1569 interception point on access denied not registered in custom interception points

nodoherty:
Just noticed that the "onAccessDenied" interception point is NOT registered in the Config.cfc

interceptorSettings.customInterceptionPoints

Added to the core Config.cfc file.


DominicWatson:
Nice one, thanks @nodoherty - released in 10.10.67

PR #629, opened 2019-05-27
PRESIDECMS-1577

john5onCheng:
Fixed upload asset error when extra meta data is set to On


sebduggan:
@john5onCheng Should we not serialize the value only if it's not a simple value? Otherwise any string value will now have quotes around it, and may break existing logic...

john5onCheng:
hi @sebduggan,

Thanks You're right. I have updated the code to ensures serialize only on non-simple value.

DominicWatson:
Nice one, thanks @john5onCheng and @sebduggan . Releasing as hotfix 10.10.68

PR #630, opened 2019-06-04
Feature presidecms 1582 formbuilder fileupload download url

chunwengpixl8:
When a user uploaded a file that has a plenty special characters in it (specifically '?'), Preside then will generate a forward slash () when converting the uploaded file to base64.

Thus, when we do the translate in the built link, Preside couldn't locate the file because of the extra forward slash generated in the URL.


DominicWatson:
Awesome, thanks for this @chunwengpixl8

PR #636, opened 2019-06-11
PRESIDECMS-1595 EnumSelect to take precedence over default field type…

vivienwong:
PRESIDECMS-1595 EnumSelect should  take precedence over default field type when generating control


DominicWatson:
Perfect, thanks @vivienwong :)

PR #638, opened 2019-06-17
PRESIDECMS-1606 added taskmanager internal handler to list of stateless URL patterns

jjannek:


DominicWatson:
Nice one, thanks @jjannek

PR #639, opened 2019-06-21
PRESIDECMS-1603 Missing asset type linkpicker quickadd quickedit form

jtcy0123:
Added asset type to validlinktypes on Linkpicker quickadd and quickedit form


DominicWatson:
Nice one, thanks @jtcy0123 :)

PR #640, opened 2019-06-25
PRESIDECMS-1608 Fix dynamic field values for configurator

sebduggan:


DominicWatson:
Brill, thanks Seb. Released in 10.10.80

PR #643, opened 2019-07-02
Auto generate friendly name from scaffold

synmal:
This will auto generate readable i18n from preside scaffold.

For example, if an object is created from Preside terminal with fields of:
1. "name"
2. "full_name"
3. "address_1"
4. "address_2"
5. "very_very_long_address_field"

It will generate as follows:
1. "Name"
2. "Full Name"
3. "Address 1"
4. "Address 2"
5. "Very Very Long Address Field"


sebduggan:
Nice enhancement, Malcolm!

May I suggest an amendment to the convert function?

```
function _convertToReadableFormat( required string words ){
	var removedNonChar = reReplace( arguments.words, "[^a-zA-Z0-9]+", " ", "all" );	
	return reReplaceNocase( removedNonChar, "\b(\w)", "\u\1", "all" );
}
```

First we add the "+" to condense multiple non-characters into a single space. But then we replace the whole of the word array looping with a simple regular expression, which is probably quicker and is much neater!

synmal:
Hi Seb,

Thanks for the suggestion. The amendments you requested already been pushed.

DominicWatson:
Super super l33t. Merging into upcoming 10.11.0 release as an enhancement.

system/services/devtools/ScaffoldingService.cfc
```diff
@@ -81,13 +81,13 @@ component singleton=true {
 		filesCreated.append( scaffoldPageTypeForm( pagetype=arguments.id, formName=arguments.id, subDir="page-types", extension=arguments.extension, fields=ListToArray( arguments.fields ) ) );
 
 
-		i18nProps[ "name" ]        = arguments.name;
-		i18nProps[ "description" ] = arguments.description;
+		i18nProps[ "name" ]        = _convertToReadableFormat( words = arguments.name );
```
DominicWatson: I think we should only be converting inputs that are field names. Descriptions and name's are already input by users as free text, not constrained. Places I believe we should convert:

1. Field names in preside objects
2. Field names in system config forms (`new configform` command)
3. Field names in widget config forms (`new widget` command)
4. Field names in page types (`new pagetype` command)

DominicWatson (changes requested):
Hey @synmal , love this PR. Could you take a look at the review feedback re which inputs we should be converting. I'll then merge into upcoming 10.11 release.

system/services/devtools/ScaffoldingService.cfc
synmal: Nice catch there. Totally overlooked the names and descriptions as free text. Changes now committed and pushed.

DominicWatson:
Nice one, thanks @synmal :) This has now been merged into the upcoming release 10.11.0. This issue will auto close when the release is finalized.

PR #644, opened 2019-07-04
PRESIDECMS-1622: Form builder post action - Include form name as email variable

choontat-pixl8:


DominicWatson:
This doesn't quite feel right @choontat-pixl8. We ought to be considering this in a bigger refactor that allows the context of a form submission to be available to an email template and supply all manner of variables in relation (rather than just the form name, which is presumably for a specific client requirement). 

Am putting this on hold for now.

PR #646, opened 2019-07-05
PRESIDECMS-1083 link picker with actvive and inactive status indication

mookeshNanda:
Added **inactive status indication** for Site tree page picker


DominicWatson:
Fantastic! Thanks @mookeshNanda - merging into upcoming 10.11.0 release

PR #647, opened 2019-07-05
Feature presidecms 1598 asset folder recursion prevention

pixl8-brayden:
Add in `currentFolderLabel` to exclude itself and children folders in results. 

In `AssetManagerService.getFoldersForSelectList()` , added an extra check and modify AJAX query string to stop the current folder and it's children from being shown in results.

Hence, user will not able to choose the folder itself and its' children folders as parent for `AssetFolderPicker` form control.

### DIFF
--- a/system/handlers/admin/AssetManager.cfc
+++ b/system/handlers/admin/AssetManager.cfc
@@ -870,9 +870,10 @@ component extends="preside.system.base.AdminHandler" {
 
 	function getFoldersForAjaxSelectControl( event, rc, prc ) {
 		var records = assetManagerService.getFoldersForSelectList(
-			  maxRows      = rc.maxRows ?: 100
-			, searchQuery  = rc.q       ?: ""
-			, ids          = ListToArray( rc.values ?: "" )
+			  maxRows            = rc.maxRows ?: 100
+			, searchQuery        = rc.q       ?: ""
+			, ids                = ListToArray( rc.values ?: "" )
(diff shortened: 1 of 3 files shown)


DominicWatson:
Hey @pixl8-brayden, this looks awesome, thank you. Is there a good reason we don't use the folder ID (rather than the label)? If this could work with explicit folder ID, that would be preferred I think (folder names are only unique within a parent folder).

pixl8-brayden:
> Hey @pixl8-brayden, this looks awesome, thank you. Is there a good reason we don't use the folder ID (rather than the label)? If this could work with explicit folder ID, that would be preferred I think (folder names are only unique within a parent folder).

Hi @DominicWatson, current approach if check with folder ID the children of current selected folder still will show in the select option. I think I missed there are cases that folder label are same, I'll try with another approach which use folder ID rather than label. Thanks. 👍 

* ps: Bryan had restructure the folder parent tree recursion, and he included this check along. If his code is fine then perhaps can close this PR. :)

pixl8-brayden:
Hi @DominicWatson,

I had update to check with folder ID instead of folder label, for prevent the folder and its child available in the parent folder selection.

Thank you
Brayden

DominicWatson:
Have merged Bryan's PR, it had some extra pieces. I liked the simplicity of this one though so was a tough call. Best work together to either pair on a ticket or make sure you don't all work on the same one's :).

PR #648, opened 2019-07-08
PRESIDECMS-1460 selectable export formbuilder

synmal:
This changes will allow users to choose which field to be included during exporting form builder submissions.

Screenshot 2019-07-08 at 10 55 38

This is done by adding a form field in formfield.xml
`<field name="export" control="yesnoswitch" default="true" sortorder="60" />`

Into the backend side, in the FormBuilderService.cfc, I added a condition which checks the formitem configuration if it is exportable or not (Refer to the changes in FormBuilderService.cfc). It will skip the formitem from being appended in the itemsToRender array if the export value is 0 or false.

### DIFF
--- a/system/forms/formbuilder/item-types/formfield.xml
+++ b/system/forms/formbuilder/item-types/formfield.xml
@@ -18,6 +18,7 @@ it will be merged with this form definition.
 			<field name="layout"    control="formbuilderFieldLayoutPicker" required="true"  sortorder="30" />
 			<field name="mandatory" control="yesnoswitch"                  required="false" sortorder="40" />
 			<field name="help"      control="textarea"                     required="false" sortorder="50" />
+			<field name="export"    control="yesnoswitch"                  required="false" sortorder="60" />
 		</fieldset>
 	</tab>
 </form>
--- a/system/forms/formbuilder/item-types/formfield.xml
+++ b/system/forms/formbuilder/item-types/formfield.xml
@@ -15,10 +15,10 @@ it will be merged with this form definition.
 					<param name="regex" value="^[a-zA-Z][a-zA-Z0-9_]*$" />
 				</rule>
 			</field>
-			<field name="layout"    control="formbuilderFieldLayoutPicker" required="true"  sortorder="30" />
-			<field name="mandatory" control="yesnoswitch"                  required="false" sortorder="40" />
-			<field name="help"      control="textarea"                     required="false" sortorder="50" />
-			<field name="export"    control="yesnoswitch"                  required="false" sortorder="60" />
+			<field name="layout"                         control="formbuilderFieldLayoutPicker" required="true"  sortorder="30" />
+			<field name="mandatory"                      control="yesnoswitch"                  required="false" sortorder="40" />
+			<field name="help"                           control="textarea"                     required="false" sortorder="50" />
+			<field binding="formbuilder_formitem.export" default="true"                                          sortorder="60" />
 		</fieldset>
 	</tab>
 </form>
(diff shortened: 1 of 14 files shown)


system/services/formbuilder/FormBuilderService.cfc
```diff
@@ -911,8 +911,9 @@ component {
 		for( var i=1; i <= formItems.len(); i++ ) {
 			if ( formItems[i].type.isFormField ) {
 				var columns = renderingService.getItemTypeExportColumns( formItems[i].type.id, formItems[i].configuration );
+				var export  = formItems[i].configuration.export ?: 0 ;
```
DominicWatson: I suggest this should be more defensive. The following approach is solid for ensuring boolean values:

```
var export = IsBoolean( formItems[ i ].configuration.export ?: "" ) && formItems[ i ].configuration.export;
```

DominicWatson: Also, only get the columns if `export = true`, so switch lines 913 + 914 around and do:

```
var columns = export ? renderingService.getItemTypeExportColumns( formItems[i].type.id, formItems[i].configuration ) : [];
```

DominicWatson (changes requested):
Hey @synmal - this looks great. Just a tiny bit of feedback to tweak. Thanks :)

system/forms/formbuilder/item-types/formfield.xml
```diff
@@ -18,6 +18,7 @@ it will be merged with this form definition.
 			<field name="layout"    control="formbuilderFieldLayoutPicker" required="true"  sortorder="30" />
 			<field name="mandatory" control="yesnoswitch"                  required="false" sortorder="40" />
 			<field name="help"      control="textarea"                     required="false" sortorder="50" />
+			<field name="export"    control="yesnoswitch"                                   sortorder="60" />
```
synmal: default="true" was removed due to the front end will always show "true" even though it was set to false. This will happen when user navigate out from the config page and reloading. Removing the default will solve the problem.

system/services/formbuilder/FormBuilderService.cfc
synmal: The changes requested had been pushed.

system/services/formbuilder/FormBuilderService.cfc
```diff
@@ -910,8 +910,8 @@ component {
 		}
 		for( var i=1; i <= formItems.len(); i++ ) {
 			if ( formItems[i].type.isFormField ) {
-				var columns = renderingService.getItemTypeExportColumns( formItems[i].type.id, formItems[i].configuration );
-				var export  = formItems[i].configuration.export ?: 0 ;
+				var export  = isBoolean( formItems[i].configuration.export ?: "" ) && formItems[i].configuration.export ?: "";
```
DominicWatson: I'm going to be really picky :) 

```
IsBoolean( somevar ?: "" ) && somevar;
// NOT:
IsBoolean( somevar ?: "" ) && somevar ?: "";
```

If `someVar ?: ""` is Boolean, then we know that it definitely exists so we don't need to the elvis operator a second time.

DominicWatson (commented):
I'm being very picky :) Have commented again. Love this PR btw, just being a git!

system/services/formbuilder/FormBuilderService.cfc
synmal: Hi @DominicWatson, I removed the extra elvis operator. Thanks for the heads up.

DominicWatson:
Have finally got round to reviewing again, thanks @synmal . Merged into upcoming 10.11.0 release :)

PR #652, opened 2019-07-09
PRESIDECMS-1143 limit notification count

mookeshNanda:
- **Limit notification count**
- **Ajax fetch notification topic as suggested by** @john5onCheng 

Default notification limit in  _config.cfc_
> settings.notificationCountLimit      = 100;

During query the countLimit will be added by 1
`select index_field from tablename where condition limit 101`

At frontend it will display as per below 
```
if ( notification.count >100 ) { // if condition meet, then add "+" after count
 display (count&"+");  // output =  "100+"
} else {
 display (count); // output will be the count, "56"
}
```

**Test result ( tested with cache = false and fwReinitCaches = true ) **
- with 20k notification ( 2 notification topic // 10.9k + 9.1k    ) 
  
`getUnreadNotificationCount`
> Total query execution time 
Before : 43.40 ms
 Now    :    1.06 ms

`getUnreadTopics`
> Total query execution time 
Before : 116.00 ms 
Now    :     4.28 ms

### DIFF
--- a/system/services/notifications/NotificationService.cfc
+++ b/system/services/notifications/NotificationService.cfc
@@ -118,7 +118,6 @@ component autodoc=true displayName="Notification Service" {
 	 */
 	public numeric function getUnreadNotificationCount(
 		  required string userId
-		, required numeric maxRows 
 	) autodoc=true {
 		var notificationCount = _getConsumerDao().selectData(
 			  filter          = { 
--- a/system/services/notifications/NotificationService.cfc
+++ b/system/services/notifications/NotificationService.cfc
@@ -121,8 +121,7 @@ component autodoc=true displayName="Notification Service" {
 		, required numeric maxRows 
 	) autodoc=true {
 		var notificationCount = _getConsumerDao().selectData(
-			  selectFields    = [ "read as notification" ]
-			, filter          = { 
+			  filter          = { 
 				  security_user = arguments.userId
 				, read          = false 
 			  }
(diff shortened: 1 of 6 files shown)


system/services/notifications/NotificationService.cfc
```diff
@@ -116,30 +116,59 @@ component autodoc=true displayName="Notification Service" {
 	 *
 	 * @userId.hint id of the admin user whose unread notification count we wish to retrieve
 	 */
-	public numeric function getUnreadNotificationCount( required string userId ) autodoc=true {
+	public numeric function getUnreadNotificationCount(
+		  required string userId
+		, required numeric maxRows 
+	) autodoc=true {
 		var queryResult = _getConsumerDao().selectData(
```
DominicWatson: This should be changed to *not use maxrows* and, instead, just do:

```
return _getConsumerDao().selectData(
      filter = {...} // full filter here, abbreviating for example
    , useCache = false
    , recordCountOnly = true // can pass this to selectData to efficiently get number of records
);
```

This avoids fetching records just to count them.

DominicWatson (changes requested):
Hey @mookeshNanda - this looks great thank you :). I've a little feedback on how the counting of notifications is working, but otherwise looks great.

mookeshNanda:
Hi @DominicWatson amended as per request 👌

system/services/notifications/NotificationService.cfc
mookeshNanda: Hi @DominicWatson amended as per request 👌

DominicWatson:
Brilliant, thanks @mookeshNanda :) Merged into 10.11.0

PR #653, opened 2019-07-09
Feature presidecms 1602 prevent raising notification for inactive cms admin

mookeshNanda:
Notification are raised only for **active** CMS admin as suggested by @nodoherty .


DominicWatson:
Brilliant, thanks @mookeshNanda :). Will be merged into the upcoming release, 10.11.0

PR #654, opened 2019-07-09
PRESIDECMS-1629

john5onCheng:
Add default value for missing mField


DominicWatson:
Nice one, thanks @john5onCheng. Will be released in 10.10.82 hotfix.

PR #655, opened 2019-07-10
Presidecms 1598 asset folder recursion

bryanlooi:
Adds an extra one-line check in `AssetManagerService.getFolderAncestors` to check if a folder has already been 'visited' to prevent infinite loops.

Refactor `AssetManagerService.getFoldersForSelectList` to include an extra argument, `excludeDescendants` which takes in a string folder ID and excludes said folder from appearing in the select list.

- Is now iterative via an array acting as a queue instead of recursive.
- Takes an extra argument, `excludeDescendants` which is a folder ID string.
- Will skip said folder when traversing down the tree effectively cutting off the sub-tree.
- Uses `selectData` only once then converts the records into a struct with folderIds as keys.
- Currently depends on selectData's `orderBy` and some special prepending and deleting to get the results in proper order.
- As an extra precaution, checks again if a certain folder has already been 'visited' to avoid duplicates.

Also adds a `one-to-many child_folders` property to the asset_folder object related to parent_folder to make it easier to access a folder's descendants rather than constantly selecting and filtering out folders.

Only sets an `excludeDescendant` when editing folders. Done with an `additionalArgs` passed into the `renderView` of `_editRecordForm` in `views/admin/assetmanager/editFolder.cfm`

### DIFF
--- a/system/handlers/admin/AssetManager.cfc
+++ b/system/handlers/admin/AssetManager.cfc
@@ -870,9 +870,10 @@ component extends="preside.system.base.AdminHandler" {
 
 	function getFoldersForAjaxSelectControl( event, rc, prc ) {
 		var records = assetManagerService.getFoldersForSelectList(
-			  maxRows      = rc.maxRows ?: 100
-			, searchQuery  = rc.q       ?: ""
-			, ids          = ListToArray( rc.values ?: "" )
+			  maxRows             = rc.maxRows ?: 100
+			, searchQuery         = rc.q       ?: ""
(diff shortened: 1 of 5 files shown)


DominicWatson:
This looks brill @bryanlooi thank you. I haven't had the time to look into it properly yet but will get merged into the upcoming release at some point soon.

DominicWatson:
Have reviewed and tested this. Works a treat. Thanks for the highly detailed work and PR notes @bryanlooi  :)

PR #656, opened 2019-07-10
Feature presidecms 1630 fixing fileuploads on using validateforms

synmal:


system/helpers/presideProxies.cfm
```diff
@@ -100,19 +100,19 @@
 	</cffunction>
 
 	<cffunction name="validateForms" access="public" returntype="any" output="false">
+		<cfargument name="formData" type="struct" required="true">
```
DominicWatson: I think this will be backward compatible breaking. Any code calling validateForms() will break with this change. What is the purpose of this change and reasoning behind it?

DominicWatson (changes requested):

system/helpers/presideProxies.cfm
synmal: In case of using multiple forms, file uploads and `validateForms()`, `validateForms()` do not convert file uploads to binary. Compared to `validateForm()`, we are required to pass in `formData` which the file uploads will be converted to binary along the way.

 However, in `validateForms()` the `formData` that it is passing to `FormsService` is declared in `PresideProxies.cfm`. So the processed `formData` stays in `validateForms()` function. One of the workarounds for this issue is to merge the forms and use `validateForm()` instead.

And regarding the change, I just realised that I had set it to `required="true"`. I will change it to `required="false"`. But I want to know your thoughts regarding this approach first.

DominicWatson: Could you try:

```cfm
<cfargument name="formData" type="struct" default="#getController().getRequestContext().getCollectionWithoutSystemVars()#" />
```

If that solves the issue, then that will also be backwards compatible.

DominicWatson: If that doesn't work, try `getCollection()` rather than `getCollectionWithoutSystemVars()` (there may be a difference with the latter doing a StructCopy() so losing reference to shared data referenced later on)

system/helpers/presideProxies.cfm
```diff
@@ -100,7 +100,7 @@
 	</cffunction>
 
 	<cffunction name="validateForms" access="public" returntype="any" output="false">
-		<cfargument name="formData" type="struct" required="true">
+		<cfargument name="formData" type="struct" default="#getController().getRequestContext().getCollection()#" />
```
synmal, line 103: Seems like `getCollection()` does fix it. Thanks for the suggestion.

DominicWatson:
Nice one thanks @synmal  - apologies for not picking this up sooner. Merging into upcoming 10.11.0 release.

PR #657, opened 2019-07-10
PRESIDECMS-1512 logout retain current page

mookeshNanda:
Redirect admin back to **previous page** if log out is done from front end.


DominicWatson:
Brilliant, thanks @mookeshNanda. MErging into the upcoming 10.11.0 release.

PR #660, opened 2019-07-16
Feature presidecms 231 domain and path next to slug field

pixl8-brayden:
Display domain and slug on the bottom of slug field.
I used JS to update `slug on display` when user typing in `slug field`. During render text input field, the checking is through `inputId`, where it is equal to `slug` then will display the domain + slug.

Thank you
Brayden

### DIFF
--- a/system/assets/js/admin/specific/siteTreePageSlugEditor/siteTreePageSlugEditor.js
+++ b/system/assets/js/admin/specific/siteTreePageSlugEditor/siteTreePageSlugEditor.js
@@ -1,0 +1,23 @@
+( function( $ ){
+	$( '.slug-editor' ).each( function(){
+		var slugEditorId = $(this).attr( 'id' );
+		$( "#slug-editor-"+slugEditorId ).html( $(this).val() );
+	});
+
+	$( '.slug-editor' ).on( 'keyup', function() {
+		var slugEditorId = $(this).attr( 'id' );
+		$( "#slug-editor-"+slugEditorId ).html( $(this).val() );
+	});
+
+	$( "input[name='parent_page']" ).on( 'change', function() {
+		$.ajax({
+			  url     : $( "input[name='parent_slug_ajax']" ).val()
+			, method  : "GET"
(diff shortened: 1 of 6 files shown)


pixl8-brayden:
presidecms 231 domain and path next to slug field screenshot

DominicWatson:
Hey @pixl8-brayden, thanks for this, and apologies for the slow time to review. This isn't quite working for me though. 

1. We should avoid "magic numbers" approaches in code where we say "if thisSpecialCircumstance then doThisSpecialThing..". In this case, using the text input form field to do something specific for a site tree slug. Instead, we should create a _specific_ form control for this purpose: e.g. `siteTreePageSlugEditor`
2. In the javascript, avoid referring to HTML element IDs where possible. This makes the field non-portable. Instead, use css classes to target the entity and ensure that the logic could work with multiple independent instances of the control in one page
3. The slug should be based on a combination of the parent page and the existing page slug. A user is able to change the parent page _and_ the slug and so the whole URL can change. I suggest detecting a change to either of these fields and posting an AJAX call to server side logic that gets the full URL.

pixl8-brayden:
Hi @DominicWatson 

Create new `siteTreePageSlugEditor` form control for link preview during edit page. AJAX call is used to get the parent page url when the parent page is changed and replaced the `parent page` part of the preview link.

CSS class selector is used and `unique id` is generated every time the form view is called, in the case of multiple target in one page JS able to target the link to replace by `id`.

Screenshot 2019-08-06 at 11 15 51 PM

Thank you
Brayden

DominicWatson:
Nice one, thanks @pixl8-brayden - there are a few tweaks to be made here, but I'll tweak them and let you know what changed. Appreciate this :)

DominicWatson:
Have merged into the upcoming release. You can see the tweaks I made here:

https://github.com/pixl8/Preside-CMS/commits/feature-PRESIDECMS-231_url-preview-next-to-site-slug

I refactored the js to make the control more portable + less jquery lookups. I also made the control work with the auto slug + used it in the add page form aswell.

PR #662, opened 2019-07-22
PRESIDECMS-1638 Setup dynamic date range options for Rules engine fi…

vivienwong:
Setup dynamic date range options for Rules engine filter


DominicWatson:
Awesome sauce @vivienwong. Merged into upcoming 10.11.0 release.

PR #663, opened 2019-07-24
PRESIDECMS-1631 fix broken multilingual og url

bryanlooi:
Updates og:url in _openGraphMeta.cfm to remove duplicate languages in url.
RegEx: `reReplace( local.ogUrl, "/([a-z][a-z]/)(\1)+", "/\1")`

Notes for, if some reason, these cases happen:
1. Does not account for capital letters e.g. '/EN/' or '/dE/' or '/Fr/'.
2. In the case there are two doubles back to back, it will not match the
   2nd one. e.g. '/de/de/de/' -> '/de/' but '/de/de/en/en/' ->
   '/de/en/en/' however '/de/de/en/fr/fr/' still becomes '/de/en/fr/'.

### DIFF
--- a/system/views/general/_openGraphMeta.cfm
+++ b/system/views/general/_openGraphMeta.cfm
@@ -6,6 +6,7 @@
 	local.title        = Trim( event.getPageProperty( "title"              ) );
 	local.mainImage    = Trim( event.getPageProperty( "main_image"         ) );
 	local.ogType       = local.site.og_type ?: "website";
+	local.ogUrl        = event.getSiteUrl() & HtmlEditFormat( event.getCurrentUrl() );
 
 	local.title  = Len( local.browserTitle ) ? local.browserTitle : local.title;
 	local.teaser = Len( local.teaser       ) ? local.teaser       : local.description;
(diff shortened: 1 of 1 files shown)


system/views/general/_openGraphMeta.cfm
```diff
@@ -14,7 +15,7 @@
 <cfoutput>
 	<meta property="og:title" content="#XmlFormat( local.title )#" />
 	<meta property="og:type"  content="#local.ogType#" />
-	<meta property="og:url"   content="#event.getSiteUrl()##HtmlEditFormat( event.getCurrentUrl() )#" />
+	<meta property="og:url"   content="#reReplace( local.ogUrl, "/([a-z][a-z]/)(\1)+", "/\1")#" />
```
DominicWatson: I think this approach is problematic:

1. It is only _patching_ the problem, and the business logic of that patch is taking place in a _view_
2. What is to say a user does not setup a structure themselves like /ca/fb/my-page.html - however unlikely that may be, when it does happen, the produced bug will have people scratching their heads

I think this should be reworked to get to the core of the problem. i.e. why is the language code in there twice. Fix that problem at the source, rather than patching it afterwards.

Does that make sense?

DominicWatson (changes requested):

system/views/general/_openGraphMeta.cfm
bryanlooi: Hi @DominicWatson,

Absolutely right. I should have mentioned that `getSiteUrl` and `getCurrentUrl` both return language slugs thus the double. I have amended it to use `event.getSiteUrl( includeLanguageSlug=false )` instead which solves the issue. There was also the option of using `getBaseUrl()` but I wasn't sure if that would break anything so I opted for the 'safer' route.

Thanks.

madmikede: Question: Is it in general correct, that `getSiteUrl` is returning the language? I think it should only return the protocol+domain+url path but not the language. Preside is redirecting to the language automatically.

DominicWatson: I think it is correct @madmikede and changing it would be compatibility breaking. As Bryan points out, you are able to override this behaviour should you wish.

DominicWatson:
Merged into upcoming 10.11.0 release. Thanks @bryanlooi

PR #665, opened 2019-07-25
PRESIDECMS-1651 Feature remove extra slash when scaffolding object

bryanlooi:
`arguments.subDir` is empty when scaffolding a basic preside object leading to the `filePath` having a double slash.


DominicWatson:
Nice one, thanks for spotting and fixing @bryanlooi :). Will be released in 10.11.0 release. I created a JIRA issue for the bug: https://presidecms.atlassian.net/browse/PRESIDECMS-1651. In future, it would be extra helpful if you could always create a ticket with any PR. Thanks!

PR #666, opened 2019-07-26
PRESIDECMS-1442 asset manager view show last uploaded

bryanlooi:
The listing table for Asset Manager now has a column for 'First uploaded' and orders the asset records by `datecreated` by default.

### DIFF
--- a/system/services/assetManager/AssetManagerService.cfc
+++ b/system/services/assetManager/AssetManagerService.cfc
@@ -471,7 +471,7 @@ component displayName="AssetManager Service" {
 	public struct function getAssetsForGridListing(
 		  numeric startRow    = 1
 		, numeric maxRows     = 10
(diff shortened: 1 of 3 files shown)


system/services/assetManager/AssetManagerService.cfc
```diff
@@ -471,7 +471,7 @@ component displayName="AssetManager Service" {
 	public struct function getAssetsForGridListing(
 		  numeric startRow    = 1
 		, numeric maxRows     = 10
-		, string  orderBy     = ""
+		, string  orderBy     = "datecreated ASC"
```
DominicWatson, line 474: This should be DESC I think :). I'll merge and change.

DominicWatson:
Nice one, thanks @bryanlooi - will make it into 10.11.0

PR #668, opened 2019-07-30
PRESIDECMS-1601 PDF thumbnail is not able to generate

pixl8-brayden:
Add checking in `NativeImageService` > `pdfPreview` function, if Lucee Server is 5 then use Java BufferedImage and Apache PDFImageWrite to generate PDF 1st page preview. 

Add new arguments `width` for `pdfPreview` function to adjust the width of PDF 1st page preview, default is set to 300.

Thank you
Brayden

### DIFF
--- a/system/services/assetManager/NativeImageService.cfc
+++ b/system/services/assetManager/NativeImageService.cfc
@@ -176,10 +176,11 @@ component displayname="Native Image Manipulation Service" {
 		,          string format
 		,          string pages
 		,          string transparent
+		,          numeric width      = 300
 	) {
 		var imagePrefix = CreateUUId();
 		var tmpFilePath = GetTempDirectory() & "/" & imagePrefix & "_page_" & arguments.page & ".jpg";
-		var allowedArgs = [ "scale", "resolution", "format", "pages", "transparent", "maxscale", "maxlength", "maxbreadth" ];
+		var allowedArgs = [ "scale", "resolution", "format", "pages", "transparent", "maxscale", "maxlength", "maxbreadth", "width" ];
 		var pdfAttributes = {
 			  action      = "thumbnail"
 			, source      = asset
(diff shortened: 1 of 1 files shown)


system/services/assetManager/NativeImageService.cfc
```diff
@@ -193,7 +194,24 @@ component displayname="Native Image Manipulation Service" {
 			}
 		}
 
-		pdf attributeCollection=pdfAttributes;
+		if ( SERVER.lucee.version.left(1) gt 4 ) {
```
DominicWatson, line 197: Extra safety checking:

```
if ( Val( Left( server.lucee.version ?: "", 1 ) ) > 4 ) {...
```

`Val()` ensures any text we have is a number valid for numeric comparison. Adding elvis here also to protect against any other unexpected changes to that variable that we can't guarantee is set (even though it isn't likely to change).

system/services/assetManager/NativeImageService.cfc
```diff
@@ -193,7 +194,24 @@ component displayname="Native Image Manipulation Service" {
...
+			tmpFilePDF  = GetTempDirectory() & imagePrefix & "_page_" & arguments.page & ".pdf";
```
DominicWatson, line 198: I think `tmpFilePDF` and `tmpFileJPG` are missing `var` scope. Will fix after merge.

system/services/assetManager/NativeImageService.cfc
```diff
@@ -193,7 +194,24 @@ component displayname="Native Image Manipulation Service" {
...
+			tmpFileJPG  = GetTempDirectory() & imagePrefix & "1.jpg";
+			FileWrite( tmpFilePDF, pdfAttributes.source );
+
+			var returnfileprefix = GetTempDirectory() & imagePrefix;
+			var BufferedImage    = createObject("java","java.awt.image.BufferedImage");
+			var ImageWriter      = createObject("java","org.apache.pdfbox.util.PDFImageWriter");
+			var document         = createObject("java","org.apache.pdfbox.pdmodel.PDDocument").load("#tmpFilePDF#");
```
DominicWatson, line 205: little thing, avoid `"#someVariable#"` if you are able to just do `someVariable` instead :) i.e.

```
...).load( tmpFile );
```

DominicWatson:
This is great, thanks @pixl8-brayden :). I've commented on a few things but I will patch those manually myself after merging your changes. Will be in the upcoming 10.11.0 release.

PR #671, opened 2019-08-06
Feature presidecms 1652 unable to select asset folder while uploading an asset

chunwengpixl8:
The query in 'getFoldersForSelectList' is selecting all the children folders as long as the parent folder is not trashed.

So, I've updated the filter, so that when it is selecting the folders for select list it will filter on the join table ( child folders ) as well.


DominicWatson:
Nice one, thanks @chunwengpixl8 - will remove the JIRA, this relates to an existing JIRA that belongs to the upcoming unreleased next version.

PR #672, opened 2019-08-14
Feature presidecms 1368 default filters

synmal:
The default filters can be defined in the `Config.cfc` like how saved filters are defined.
Example: 
<pre><code>settings.filters = {
      publishedStuff = { filter = "stuff.published=true" }
    , approvedStuff = { filter = "stuff.approved=true" }
};</code></pre>

To use the saved filters as a default filter, it must be defined in the `Object.cfc`.
<pre><code>/**
 * @datamanagerGridFields label,approved,published
 * @defaultFilters publishedStuff,approvedStuff
 */
component dataManagerGroup="Lookup" {
	property name="approved" type="boolean" dbtype="boolean";
	property name="published" type="boolean" dbtype="boolean";
}</code></pre>

So whenever, `selectData()` is called, it will automatically filter the query using the default filters.

These default filters can be ignored by specifying which filter to be ignored.
<pre><code>args.stuff = presideObjectService.selectData(
      objectName = "stuff"
    , selectFields = [
          "label"
        , "approved"
	, "published"
    ]
    , ignoreDefaultFilters = [ "publishedStuff" ]
);</code></pre>

How it works is by just appending the default filters( the ones that are not ignored ) to the `arguments.savedFilters` in the `PresideObjectService.cfc`. I had also added a condition in `_getDefaultFilters()` private function to check if the current request is an admin request because without it, default filters will be applied to the data manager as well.

### DIFF
--- a/system/services/presideObjects/PresideObjectService.cfc
+++ b/system/services/presideObjects/PresideObjectService.cfc
@@ -192,14 +192,18 @@ component displayName="Preside Object Service" {
 		,          array   bypassTenants           = []
 		,          array   ignoreDefaultFilters    = []
 	) autodoc=true {
-		var defaultFilters = listToArray( getObjectAttribute( arguments.objectName, "defaultFilters", "" ) );
-		if( !defaultFilters.isEmpty() ){
-			if( !arguments.ignoreDefaultFilters.isEmpty() ){
-				arguments.ignoreDefaultFilters.each( function(element){
-					defaultFilters.delete(element);
-				});
+		var adminRequest = $getRequestContext().isAdminRequest();
+		
+		if ( !adminRequest ){
+			var defaultFilters = listToArray( getObjectAttribute( arguments.objectName, "defaultFilters", "" ) );
+			if( !defaultFilters.isEmpty() ){
+				if( !arguments.ignoreDefaultFilters.isEmpty() ){
+					arguments.ignoreDefaultFilters.each( function(element){
+						defaultFilters.delete(element);
+					});
+				}
+				arguments.savedFilters.append( defaultFilters, true );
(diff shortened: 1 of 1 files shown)


system/services/presideObjects/PresideObjectService.cfc
```diff
@@ -2947,6 +2951,25 @@ component displayName="Preside Object Service" {
 		return expanded;
 	}
 
+	private array function _getDefaultFilters( required string objectName, required array ignoreDefaultFilters ){
+		var adminRequest = $getRequestContext().isAdminRequest();
+		var defaultFilters = [];
+
+		if ( !adminRequest ){
```
DominicWatson: We should not assume that the default filters are not to be used in admin requests. For example, we may have an admin only application that wants to use default filters. 

If more fine-grained default filters want to be used, interceptors can already be used to achieve this.

DominicWatson (changes requested):
This looks pretty much perfect thanks @synmal . Just one review comment re not using the filters in the admin (they should be).

system/services/presideObjects/PresideObjectService.cfc
synmal: Hi @DominicWatson.

I had removed the admin checking. Please do let me know if there is anything else.

Thank you,
Malcolm

DominicWatson:
Nice one, thanks @synmal. Merged into upcoming release 10.11.0 :)

DominicWatson:
@synmal it would be really awesome if you could add some documentation for this here: https://github.com/pixl8/Preside-Documentation/blob/develop/docs/02.devguides/02.dataobjects/page.md

PR #673, opened 2019-08-21
PRESIDECMS-1365 time taken on tasks screen

bryanlooi:
Just added a new column on `views/admin/taskmanager/index` table for time taken. Only shows if `task.was_last_run_success` is true else is blank.


DominicWatson:
Fab, thanks @bryanlooi . merging into upcoming release branch.

PR #674, opened 2019-08-28
Feature presidecms 1663 data too long for column ip_address

nodoherty:


DominicWatson:
Nice one, thanks @nodoherty . Released in 10.10.89

PR #675, opened 2019-08-30
PRESIDECMS-1557

john5onCheng:
Hi @DominicWatson ,

I just got chance to work on this bit again. I found a missing bit which is to skip validation from validationEngine if supressfield is available.


DominicWatson:
Fab, thanks @john5onCheng - merged.

PR #676, opened 2019-08-30
Feature presidecms 1656 email overwrite domain

jjannek:
Added the option to overwrite emails for buildLink. Implicitely other RequestContectDecorator functions need to use this as well, e.g. getSiteUrl - because of custom route handlers making use of those.

In addition there is a new email layout option for the default layout to set a custom domain that overwrite the default one within email center preview and during actual sending.


jjannek:
I just made sure that the test suite does not crash. But it could be good adding corresponding tests for this functionality. Could do that later once we agree on whether the actual implementation is fine as is.

DominicWatson:
This is great, merging into upcoming 10.11.0 release. Would be great to add some docs into the develop branch of https://github.com/pixl8/Preside-Documentation

PR #678, opened 2019-09-11
Make sure the subscriber get the selectfields required

annbelievable:
This is to ensure the subscriber object select the field 'get_email_notifications' which is needed or else the email will not be sent out.


DominicWatson:
Nice one, thanks @annbelievable . Merged into 10.11.0 (caused by PRESIDECMS-1602 that was part of 10.11.0).

PR #683, opened 2019-09-18
PRESIDECMS-1679 email centre versioning bug settings tab

nodoherty:
Added some code to simply ensure the Settings tab abides by the draft state of the email template being edited. This should hopefully ensure the template notices align and the action buttons.


DominicWatson:
Brill, thanks Niall. Released in 10.10.94.

PR #684, opened 2019-09-19
PRESIDECMS-1681 Apply control based on menu context

vivienwong:
Apply exclude_children_from_navigation check to sub menu, instead of sub nav check


DominicWatson:
Nice one, thanks for this @vivienwong . Released in 10.10.94.

PR #685, opened 2019-09-19
PRESIDECMS-1682 Option to include pageIconCategory for brands icon

vivienwong:
Setup option for additional fontawesome icon class, e.g. brand icons requires "b" in addition to standard "fa" in order to display correctly


DominicWatson:
Discussed together, not quite right. There is an easier way :)

PR #686, opened 2019-09-19
PRESIDECMS-1682 Remove lines that break brands icons

vivienwong:
Removed classes or lines that broke brand icons


DominicWatson:
Awesome :) Pun intended. Thanks @vivienwong - merged into 10.11.0.

PR #687, opened 2019-09-23
PRESIDECMS-1688 Include new field: datecreated for grid listing display

vivienwong:
Include datecreated field for trash folder grid listing


DominicWatson:
Nice one, thanks @vivienwong. Merging now.

PR #688, opened 2019-09-23
PRESIDECMS-1684 new getitems function for enum service

pixl8-brayden:
New functions for Enum Service

- getStructItems( required string enum, string value = "" )
return struct of the enum detail. if no value provided return all enum values in struct, if string / list of values provided, return found enum values in struct.

- getLabelByValue( required string enum, required string value = "" )
return label of the provided enum value, return empty string if not found.

- getValueByLabel( required string enum, required string label = "" )
return value of the provided enum label, return empty string if not found.

### DIFF
--- a/system/services/enum/EnumService.cfc
+++ b/system/services/enum/EnumService.cfc
@@ -45,6 +45,62 @@ component {
 		return itemsWithLabelsAndDescriptions;
 	}
 
+	public struct function getStructItems( required string enum, string value = "" ) {
+		var enums                          = _getConfiguredEnums();
+		var rawItems                       = enums[ arguments.enum ] ?: [];
+		var value                          = arguments.value         ?: "";
+		var itemsWithLabelsAndDescriptions = {};
+		var valueEnumResult                = {};
+
+		for( var itemId in rawItems ) {
+			itemsWithLabelsAndDescriptions[ itemId ] = {
+				  label       = $translateResource( uri="enum.#arguments.enum#:#itemId#.label"      , defaultValue=itemId )
+				, description = $translateResource( uri="enum.#arguments.enum#:#itemId#.description", defaultValue=""     )
+			}
+		}
+
+		if ( !isEmpty( arguments.value ) ) {
+			for ( item in listToArray( value, "," ) ) {
+				if ( itemsWithLabelsAndDescriptions.keyExists( item ) ) {
+					valueEnumResult[ item ] = itemsWithLabelsAndDescriptions[ item ];
+				}
+			}
+
+			return valueEnumResult;
+		} else {
+			return itemsWithLabelsAndDescriptions;
+		}
+	}
+
+	public string function getLabelByValue( required string enum, required string value = "" ) {
+		var enums    = _getConfiguredEnums();
+		var rawItems = enums[ arguments.enum ] ?: [];
+		var value    = arguments.value         ?: "";
+
+		for( var itemId in rawItems ) {
+			if ( itemId eq value ) {
+				return $translateResource( uri="enum.#arguments.enum#:#itemId#.label" , defaultValue=itemId );
+			}
+		}
+
+		return "";
+	}
+
+	public string function getValueByLabel( required string enum, required string label = "" ) {
+		var enums    = _getConfiguredEnums();
+		var rawItems = enums[ arguments.enum ] ?: [];
+		var label    = arguments.label         ?: "";
+
+		for( var itemId in rawItems ) {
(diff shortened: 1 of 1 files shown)


system/services/enum/EnumService.cfc
```diff
@@ -45,6 +45,62 @@ component {
...
+
+		for( var itemId in rawItems ) {
+			itemsWithLabelsAndDescriptions[ itemId ] = {
+				  label       = $translateResource( uri="enum.#arguments.enum#:#itemId#.label"      , defaultValue=itemId )
+				, description = $translateResource( uri="enum.#arguments.enum#:#itemId#.description", defaultValue=""     )
+			}
+		}
+
+		if ( !isEmpty( arguments.value ) ) {
+			for ( item in listToArray( value, "," ) ) {
```
DominicWatson, line 63: Missing `var` here for the `item` variable. But actually, this loop and condition on 62 could be removed (see comment above).

system/services/enum/EnumService.cfc
```diff
@@ -45,6 +45,62 @@ component {
 		return itemsWithLabelsAndDescriptions;
 	}
 
+	public struct function getStructItems( required string enum, string value = "" ) {
```
DominicWatson, line 48: I like this idea, could we rename the method, and rejig the signature though:

```cfc
public struct function getEnumProperties( required string enum, array restrictToKeys=[], array properties=[ "label", "description" )
```

The properties can be dynamically used to loop over in the `translateResource()` calls. This way calling code can customize implementations, i.e. add `iconClass`, etc.

Also, note the preferred use of an `array` type, rather than using string lists.

system/services/enum/EnumService.cfc
```diff
@@ -45,6 +45,62 @@ component {
...
+		var enums                          = _getConfiguredEnums();
+		var rawItems                       = enums[ arguments.enum ] ?: [];
+		var value                          = arguments.value         ?: "";
+		var itemsWithLabelsAndDescriptions = {};
+		var valueEnumResult                = {};
+
+		for( var itemId in rawItems ) {
+			itemsWithLabelsAndDescriptions[ itemId ] = {
```
DominicWatson, line 56: Could simplify this, and not need to loop twice, by adding:

```cfc
if ( isEmpty( arguments.restrictToKeys ) || ArrayFindNoCase( arguments.restrictToKeys, itemId ) ) {
// ... add to struct that will be returned
}
```

system/services/enum/EnumService.cfc
```diff
@@ -45,6 +45,62 @@ component {
...
+		var rawItems = enums[ arguments.enum ] ?: [];
+		var value    = arguments.value         ?: "";
+
+		for( var itemId in rawItems ) {
+			if ( itemId eq value ) {
+				return $translateResource( uri="enum.#arguments.enum#:#itemId#.label" , defaultValue=itemId );
+			}
+		}
+
+		return "";
```
DominicWatson, line 86: I think all of this logic can be replaced with:

```cfc
return $translateResource( uri="enum.#arguments.enum#:#arguments.value#.label", defaultValue=arguments.value );
```

system/services/enum/EnumService.cfc
```diff
@@ -45,6 +45,62 @@ component {
...
+	}
+
+	public string function getValueByLabel( required string enum, required string label = "" ) {
+		var enums    = _getConfiguredEnums();
+		var rawItems = enums[ arguments.enum ] ?: [];
+		var label    = arguments.label         ?: "";
+
+		for( var itemId in rawItems ) {
+			var itemLabel = $translateResource( uri="enum.#arguments.enum#:#itemId#.label" , defaultValue=itemId );
```
DominicWatson, line 95: After refactoring `getLabelByValue` as above, should re-use that method here:

```cfc
var itemLabel = getLabelByValue( arguments.enum, itemId );
```

DominicWatson (changes requested):
Hey @pixl8-brayden - thanks for this. I've a few review comments that would be good to look at before we merge this. Apologies for slow uptake on getting to this - if we can get this ready Monday can make it into 10.11.0 release :)

DominicWatson:
@pixl8-brayden I went ahead and merged this into 10.11.0 and made suggested changes. The methods are now:

```cfc
public struct function getEnumProperties( 
	  required string enum
	,          array  restrictToKeys = [],
	,          array  properties     = [ "label", "description" ] 
) {...}

public string function getLabelByKey( 
	  required string enum
	, required string key 
) {...}

public string function getKeyByLabel( 
	  required string enum
	, required string label 
) {...}
```

pixl8-brayden:
Hi @DominicWatson thanks for the suggestion and the modification. Glad to heard this make it to 10.11 :)

PR #689, opened 2019-09-23
PRESIDECMS-1689 Check presence of sourceobjectname before proceeding …

vivienwong:
…with condition


DominicWatson:
Nice one, thanks @vivienwong . Will be released with 10.11.0.

PR #691, opened 2019-09-24
PRESIDECMS-1690 asset manager table layout break

pixl8-brayden:
Expected outcome as per screenshot
asset manager table screenshot

Thank you
Brayden


DominicWatson:
Fab, thanks @pixl8-brayden - merged into 10.11.0 :)

PR #692, opened 2019-09-24
Feature presidecms 1683 anonymous user tracking is not working when full page caching is enabled

pixl8-brayden:
Added `websiteUserActionService.recordAction(...)` in PageCachingInterceptor. When local cached is found it still recordAction for page visit.

Thanks
Brayden


DominicWatson:
Brill, nice one thanks @pixl8-brayden . I tweaked slightly, used `event.getCurrentPageId()` to get the page ID so as not to need DB lookup. Also, only called the logging service when pageId has length.

PR #693, opened 2019-09-26
PRESIDECMS-1686 svg derivative support

mookeshNanda:
Hi @DominicWatson 

## SVG derivative support
When derivates are used for SVG file type, it will be converted to png.
- Support both ImageMagickService and NativeImageService
- Added Java Batik Library for NativeImageService.

**Thanks & Regards**
Mookesh Nanda


DominicWatson:
Hey @mookeshNanda - this is brilliant, thank you. There are a lot of asset manager changes in 10.11.0 and I've had to take what you've done and change the shape of it a lot. I also found that the imagemagick conversion did not always work with transparency and I couldn't get it to work. So I reshuffled to always use the java approach for SVG conversion (for now).

Would not have got it done without this PR though, so thank you.

DominicWatson:
Somehow I messed up with my merging which means I need to close this PR. The changes are in though :+1: Apologies for them not being correctly attributed.

PR #694, opened 2019-10-01
PRESIDECMS-1697 emai centre scheduling settings issue

nodoherty:
Because the getTemplate() method wasn't returning the most up-to-date version of the template this was causing issues when modifying the settings. So updated to always select the data from the database directly rather than cache.


DominicWatson:
Nice one, thanks @nodoherty - released on its own in 10.10.96.

PR #696, opened 2019-10-16
PRESIDECMS-1705 added missing slug validator for asset slug

mookeshNanda:
Added missing slug validator for asset's filename slug.


DominicWatson:
Hi @mookeshNanda I'm not sure on this one. Can you confirm, with the latest 10.11 release that you are able to add mixed case file names? Also, does the system break if you choose a filename that does not conform to a slug format?

mookeshNanda:
**Hi @DominicWatson ,**

Have added **fileNameSlug** into validator. It accept alphanumeric( lower & uppercase ) and dashes.

**Also, does the system break if you choose a filename that does not conform to a slug format?** Nope, it doesn't break. It seems, it strip out unwanted chars. 

Let me know, if you have any suggestion

**Thanks & Regards**
Mookesh Nanda

DominicWatson:
Great, thanks @mookeshNanda - merging into upcoming release now.

PR #698, opened 2019-10-21
Feature preside 1711 delete object with fullstop character

annbelievable:
Select an object's labelfield as 'label' then getting the labelfield from the object using the word 'label' too


DominicWatson:
Nice one thanks @annbelievable :). An alternative would be to use a formula field here:

```cfc
property name="label" formula="${prefix}some_relationship.label";
```

PR #699, opened 2019-10-21
PRESIDECMS-1712 Do not allow editing of blueprint recipient type

sebduggan:


DominicWatson:
Love it, thanks @sebduggan

PR #700, opened 2019-10-23
New devtools command to manage SSL certificates

smoufa:
Devtools command to list and import SSL certificates. Useful if Lucee admin is disabled.


DominicWatson:
Super nice! I suggest however you release this as an extension? It shouldn't have to wait for a next Preside release to be useful + not really something for a hotfix.

I can help setup the travis build and publish part :+1:

PR #702, opened 2019-10-30
PRESIDECMS-1719 runsql function arguments issues

nelsonchuah:
Hi Dom,

Could you help review the code below.

Issue: 
runSql function is updating params argument so this will cause an issue when we try to use the same params twice

Thank you
Nelson

### DIFF
--- a/system/services/database/SqlRunner.cfc
+++ b/system/services/database/SqlRunner.cfc
@@ -29,7 +29,8 @@ component singleton=true {
 		}
 
 		if ( StructKeyExists( arguments, "params" ) ) {
-			for( var param in arguments.params ){
+			var argumentsParams = duplicate( arguments.params );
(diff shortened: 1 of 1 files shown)


DominicWatson:
Hey @nelsonchuah - thanks for digging into this one.

I think we may be able to be more accurate here and avoid doing `Duplicate()`. On line 58, where the deletion of the name field happens, we could instead do something like:

```cfc
params[ param.name ] = {
    value     = param.value     ?: NullValue()
  , cfsqltype = param.cfsqltype ?: NullValue()
  , // .. and so on
};
```

If we cover all the actual allowed attributes on cfqueryparam tag, I think we should be good (not sure about the NullValue() bit - this is off the top of my head).

DominicWatson:
Closing this for now as alternative approaches were found.

PR #703, opened 2019-11-07
PRESIDECMS-1723 assetmanager prehandler update

annbelievable:
The action `uploadAssetAction` have rc.asset_folder and did not have rc.folder. When checking for permission, it will always check for root folder permission instead of the folder that an admin is trying to upload the asset to (rc.asset_folder). I have added a check so that it checks if rc.asset_folder is present, if it is present, then it will set it to rc.folder. Then the logic will check if rc.folder is present, else set it to root folder.


DominicWatson:
Hey @annbelievable - looks good :+1: Will get this merged asap

PR #704, opened 2019-11-07
Feature presidecms 1725 remove extraneous character

nodoherty:
Character removed


DominicWatson:
Nice one thanks @nodoherty  - not sure how that made it in!

Will be in 10.10.6 hotfix.

PR #705, opened 2019-11-08
Feature presidecms 1726 formbuilder export default

synmal:
Changed the formbuilder export to be true by default.


DominicWatson:
Hi @synmal , thanks for this. It doesn't quite address the problem though. Instead, the logic should be _reversed_. i.e. a user should choose to _exclude_ an item from the export. 

The point here is that all existing data for those upgrading will not have this field set. For backward compatibility, all those configured fields should remain as downloadable.

So, field should be renamed to exclude... rather than include and the logic changed to leave those fields flagged with the exclude field out of the export.

synmal:
Hi @DominicWatson, the logic for exporting fields has been reversed in the latest change.

PR #706, opened 2019-11-08
PRESIDECMS-1727 enable Uberselect control for mobile

tuandat3019:
fix bug can't view,add,edit preside object table in mobile


DominicWatson:
Ha, nice one thank you. This code was taken from another control some time ago. Good to know mobile browsers are caught up :)

PR #708, opened 2019-11-08
PRESIDECMS-1729 fix auto size column for excel

tuandat3019:
image
this will never work because headers always have len = 0
i change it to dataCols


DominicWatson:
Brill :) Nice find thank you! Will be released in 10.11.6

PR #709, opened 2019-11-08
PRESIDECMS-1730 Multi Tenancy

tuandat3019:
Use more than 1 tenant in preside object
image


DominicWatson:
Hey thanks for this @tuandat3019 . Looks great. We'll definitiely need to fix any failing tests + create some new tests to ensure this works as expected (and continues to work in the future). Hopefully will make it into 10.12.

PR #710, opened 2019-11-08
PRESIDECMS-1731 Quick form for preside object

tuandat3019:
image

use anotation @quickform true in preside object to enable this feature


DominicWatson:
Hey thanks  for this @tuandat3019 . Will need a good look at this before merging. But definitiely looks linteresting.

PR #711, opened 2019-11-08
PRESIDECMS-1734 Fix latest version issue with versionNavigator

sebduggan:
Incorrect when there is no previous version


DominicWatson:
Nice one, thanks @sebduggan  - going into 10.11.8

PR #715, opened 2019-11-13
PRESIDECMS-1722 identified potential race conditions missing var

pixl8-brayden:
Add missing var, but for some of the cases I not so sure about it then I leave it as it is.
Thanks

### DIFF
--- a/system/handlers/admin/DataManager.cfc
+++ b/system/handlers/admin/DataManager.cfc
@@ -2198,7 +2198,7 @@ component extends="preside.system.base.AdminHandler" {
 		}
 
 		if ( presideObjectService.deleteData( objectName=object, filter={ id = ids } ) ) {
-			for( var record in records ) {
+			for( record in records ) {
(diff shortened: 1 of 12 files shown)


system/handlers/admin/DataManager.cfc
```diff
@@ -2198,7 +2198,7 @@ component extends="preside.system.base.AdminHandler" {
 		}
 
 		if ( presideObjectService.deleteData( objectName=object, filter={ id = ids } ) ) {
-			for( record in records ) {
+			for( var record in records ) {
```
DominicWatson: This one is unecessary - var'd earlier in the function

DominicWatson (commented):
Thanks for this @pixl8-brayden. See review comments.

system/handlers/admin/DataManager.cfc
pixl8-brayden: I must missed it, I had removed the unnecessary var. Thanks

PR #716, opened 2019-11-18
Fix: Formbuilder export icon missing.

seakchiew:
minor tweak on formbuilder layout where export icon is not visible due to `fa-inverse` class on a white background


DominicWatson:
Thanks @seakchiew - you were a few minutes too late though! See https://www.preside.org/release-notes/release-notes-for-10-11-9.html.

Closing (as already fixed)

PR #718, opened 2019-11-25
PRESIDECMS-1755 sort extensions function loop infinitely

jtcy0123:
Updated _sortExtensions function logic when sorting dependant extensions to move the extensions they are depending on that lies behind them up to a position above them instead of swapping their position.


DominicWatson:
Brilliant, thanks @jtcy0123 . Will release in its own hotfix.

DominicWatson:
Released in 10.11.16, really appreciate the time digging into this.

PR #719, opened 2019-11-28
PRESIDECMS-1654 asset url is too long

choontat-pixl8:
Extend `asset_url` field length to 1024 and remove unique indexes for `asset_url` fields.


DominicWatson:
Thanks for this @choontat-pixl8 . I'm weary of pulling this into a hotfix as it will make index changes on potentially large db tables. But I'm not ignoring :). Will pull into upcoming release at some point.

john5onCheng:
Instead of increasing the length for asset url. Perhaps we can reduce the length of filename field and remove the redundant path e.g. https://www.abc.com/uploads/assets/ or just remove the domain to reduce the length

DominicWatson:
The domain part, etc. is not redundant. You could potentially change this and have some assets still serving on a different domain, for example.

I have merged into upcoming 10.12.0 snapshot.

Thanks for this @choontat-pixl8

PR #720, opened 2019-12-02
PRESIDECMS-1757

john5onCheng:
No derivative created when versioned image / crop / focal point is updated or added


DominicWatson:
:+1: great stuff, thanks @john5onCheng - will be released in 10.11.17

PR #721, opened 2019-12-06
Datamanager export permission customisation

jjannek:
added params to allow data manager customisations for the allowDataExport flag. This is useful if a custom website wants to distinguish export and read permissions.
I did not want to introduce an export permission in the core as this would break a lot of existing applications. This way now gives 2 possibilities to use a custom export permission, either by evaluating a permission in preRenderListing or by customising checkPermission.

### DIFF
--- a/system/handlers/admin/DataManager.cfc
+++ b/system/handlers/admin/DataManager.cfc
@@ -64,7 +64,7 @@ component extends="preside.system.base.AdminHandler" {
 				, isMultilingual      = IsTrue( prc.isMultilingual  ?: "" )
 				, draftsEnabled       = IsTrue( prc.draftsEnabled   ?: "" )
 				, canDelete           = IsTrue( prc.canDelete       ?: "" )
-				, allowDataExport     = IsTrue( prc.allowDataExport ?: "" )
+				, allowDataExport     = structKeyExists( prc, "allowDataExport" ) ? IsTrue( prc.allowDataExport ?: "" ) : true
 			}
 		);
 	}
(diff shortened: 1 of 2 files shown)


DominicWatson:
Hey @jjannek - could you give an overview of how a dev would use this to control, per object, data export access?

jjannek:
The allowDataExport flag can now be in the prc scope, so one solution to add custom handling to this would be to evaluate and set it before the listing view gets generated, e.g. by using the preRenderListing customization. A developer can define whatever permission for the export she wants and set it accordingly. For example (within the object related datamanager handler, e.g. `/handlers/admin/datamanager/myobject.cfc`):

        private string function preRenderListing( event, rc, prc, args={} ) {
		
		prc.allowDataExport = hasCmsPermission( "datamanager.export" );
		
		return "";
	}

The `datamanager.export` permission is custom and could be anything you want as a developer. You just need to define it in Config.cfc, add it to a role, and evaluate it in code.
An alternative approach could be to customize `checkPermission`, but that would of course mean customizing all of the permission checking for an object which might be more complex than the above solution.

DominicWatson:
Right - ok. I wonder whether we could create a couple of object annotations for this aswell/instead:

```cfc
/**
 * @dataManagerExportEnabled false (default true)
 * @dataManagerExportPermissionKey export (default read)
 */
component {
 // ..
}
```

You would then add methods to `DataManagerService.cfc`: `isDataExportEnabled( objectName )`, `getDataExportPermissionKey( objectName )`. 

I think that would then offer clean interface to achieve this.

jjannek:
Yep, that would be awesome. I was looking for a straight forward approach to just make it work - with the least amount of code changes as possible. ;-)
Your proposed solution is of course far better.

DominicWatson:
Do you think you could squeeze that in and test locally before we merge this? Would rather have that in one pull rather than leave it till later if thats ok?

DominicWatson:
Looks perfect, merging into 10.11.18. Thanks Jan.

PR #722, opened 2019-12-09
PRESIDECMS-1761 Setup value for placeholder option, switch to hide pl…

vivienwong:
PRESIDECMS-1761 Setup value for placeholder option, switch to hide placeholder


DominicWatson:
Nice one thanks @vivienwong - releasing in 10.11.19

PR #724, opened 2019-12-16
PRESIDECMS-1708 allow vid expiry configurable

linfeng2011:
Presidecms 1708 allow vid cookie expiry configurable per application


DominicWatson:
Nice one, thanks @linfeng2011. Merging into 10.11.22.

PR #727, opened 2019-12-31
PRESIDECMS-1770 unable save draft field for many to many relationship

pixl8-brayden:
Line 202 - 208 added for select live content only or with draft content in many to many relationship table.
Line 420 and 708 added to pass the isDraft argument to syncManyToManyData function.

Update in syncManyToManyData function to check existing record and pass in correct isDraft value when inserting record.

One known issue from this PR, after save as draft it working perfectly, but when publish the draft the object versioning won't get updated due to the length of the of many to many field still the same and it consider as no change field (VersioningService.cfc line 306). Need some help from someone familiar with how the versioning work and how to solve this.

Thanks
Brayden

### DIFF
--- a/system/services/presideObjects/PresideObjectService.cfc
+++ b/system/services/presideObjects/PresideObjectService.cfc
@@ -199,6 +199,13 @@ component displayName="Preside Object Service" {
 
 		if ( !args.allowDraftVersions && !args.fromVersionTable && objectIsVersioned( args.objectName ) ) {
 			args.extraFilters.append( _getDraftExclusionFilter( args.objectname ) );
+
+			if ( ( arguments.selectManyToMany ?: false ) and !isEmpty( arguments.relationshipTable ?: "" ) and !$getRequestContext().isAdminRequest() ) {
+				args.extraFilters.append( {
+					  filter       = "#arguments.relationshipTable#._version_is_draft is null or #arguments.relationshipTable#._version_is_draft = :#arguments.relationshipTable#._version_is_draft"
+					, filterparams = { "#arguments.relationshipTable#._version_is_draft"=false }
+				} );
+			}
 		}
 
 		args.extraFilters.append( _expandSavedFilters( argumentCollection=arguments ), true );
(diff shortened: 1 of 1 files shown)


system/services/presideObjects/PresideObjectService.cfc
```diff
@@ -199,6 +199,13 @@ component displayName="Preside Object Service" {
 
 		if ( !args.allowDraftVersions && !args.fromVersionTable && objectIsVersioned( args.objectName ) ) {
 			args.extraFilters.append( _getDraftExclusionFilter( args.objectname ) );
+
+			if ( ( arguments.selectManyToMany ?: false ) and !isEmpty( arguments.relationshipTable ?: "" ) and !$getRequestContext().isAdminRequest() ) {
+				args.extraFilters.append( {
```
DominicWatson, line 204: I think this line can and should do:

```cfc
args.extraFilters.append( _getDraftExclusionFilter( arguments.relationshipTable ) );
```

?

system/services/presideObjects/PresideObjectService.cfc
```diff
@@ -199,6 +199,13 @@ component displayName="Preside Object Service" {
 
 		if ( !args.allowDraftVersions && !args.fromVersionTable && objectIsVersioned( args.objectName ) ) {
 			args.extraFilters.append( _getDraftExclusionFilter( args.objectname ) );
+
+			if ( ( arguments.selectManyToMany ?: false ) and !isEmpty( arguments.relationshipTable ?: "" ) and !$getRequestContext().isAdminRequest() ) {
```
DominicWatson, line 203: What is the purpose of only doing this for non-admin requests? We already have the `!args.allowDraftVersions` check in the containing `if` condition. Is this not sufficient?

DominicWatson (commented):
Hi @pixl8-brayden, this looks great - though clearly contains risk as a change so would like to make sure we review and test this thoroughly. I have made a couple of comments, could you review?

Also, how much testing has been done on this? Do we have a reproduceable case we can try out?

DominicWatson:
Closing now. Superceded by KokAnn's pr. Thanks for all the work going into both.

PR #729, opened 2020-01-08
Support for MSSQL varchar(max)

jjannek:


DominicWatson:
From discussion, my preference would be that a setting in Config.cfc would control whether or not the adapter chose `varchar(max)` instead of `text`, _always_. So:

```settings.mssql.useVarcharMaxForText = true;```

Set that and you are done. No need to override core/extension objects to change the db type on them.

This setting could be passed into the AdapterFactory and then passed into the MSSQL DB adapters so that they can then choose `varchar(max)` or `text` when evaluating the DB type mappings.

_This setting would need to default to **false** for backwards compatibility._

sebduggan:
Thanks @jjannek - merged into v10.11.24

PR #733, opened 2020-01-20
PRESIDECMS-1767 Setup option to expand super quick add filter form modal box

vivienwong:
PRESIDECMS-1767 Setup option to expand super quick add filter form modal box


DominicWatson:
Nice one, thanks @vivienwong. Will be released in 10.11.25

PR #735, opened 2020-01-31
Feature presidecms 1776 asset items overlapping asset name on asset manager

pixl8-brayden:
Limit the css of asset image max-width to 100%, to avoid overlapping happened in smaller screen size.


DominicWatson:
Nice one, thanks @pixl8-brayden - released in 10.11.28

PR #737, opened 2020-02-04
PRESIDECMS-1791 Prioritise args.footerEnable false value if available

vivienwong:
PRESIDECMS-1791 Prioritise args.footerEnable false value if available


DominicWatson:
Nice one, thanks @vivienwong - being released in 10.11.28

PR #739, opened 2020-02-17
PRESIDECMS-1795

john5onCheng:
Add option to hide top level sub sibling menu


DominicWatson:
Hey @john5onCheng thanks for this. Could you expand on what this does, perhaps provide an annotated screenshot? I'm not clear what is happening here (I'm sure it's good, just want to understand it clearly before releasing :).

john5onCheng:
Hi Dom,

Here is the use case with screenshot
https://presidecms.atlassian.net/browse/PRESIDECMS-1795

Many thanks
J

PR #740, opened 2020-02-20
PRESIDECMS-1797 remove case sensitive when search enum key by label

pixl8-brayden:
Always compare lower case between enum label with provided label.


DominicWatson:
Hi @pixl8-brayden thanks for this. I'm curious though, CFML `==` comparison for strings is not case sensitive. Was the original code not matching when the two strings had differing case but the same letters?

pixl8-brayden:
Hi @DominicWatson 
Yes, was faced this in one of project and comparing enum label with value in all caps didn't get expected match. So I was thinking just to be safe and convert to lower case before comparing.
Thanks

PR #741, opened 2020-02-21
PRESIDECMS-1798 Read UTF-8 from i18n files

tuandat3019:
support utf-8 character for .properties files
you can write utf-8 format : 

"Đạt đẹp trai" 

instead of convert to 

"\u0110\u1EA1t \u0111\u1EB9p trai".

I think it will save a lot of time for countries using the UTF-8 format

It works well with the old converted utf-8 format


DominicWatson:
For whatever reason, the "standard" format for .properties files is not UTF-8. The tooling for translating using them expects them to be in the encoding that we are using (at least they used to). There are, for example, plugins for Preside that will auto encode and decode for you so that you don't need to worry about the encoding. 

Closing this for now - I'd rather not risk breaking things with zanata, etc.

Thank you for the Pull though.

PR #744, opened 2020-02-27
PRESIDECMS-1802 Select distinct ID to prevent duplicate from version

vivienwong:
PRESIDECMS-1802 Select distinct ID to prevent duplicate from version


DominicWatson:
Nice one, thanks @vivienwong . FYI - there is a `distinct = true` argument that can be passed to selectData(). Using that instead.

PR #745, opened 2020-03-04
Feature presidecms 1805 data export many to one sort order

nelsonchuah:
Hi Dom,

There's some issue with the recently added sorting for excel export for fields that are many-to-one

have updated to check if fields are many-to-one to user object's labelfield

Thank you
Nelson


DominicWatson:
Perfect, thanks @nelsonchuah

PR #747, opened 2020-03-12
PRESIDECMS-1808

john5onCheng:
postLoginUrl is not picking up rc scope data, it keep getting from session storage which is not correct

### DIFF
--- a/system/handlers/Login.cfc
+++ b/system/handlers/Login.cfc
@@ -12,7 +12,7 @@ component {
 		}
 		var loginId      = rc.loginId  ?: "";
 		var password     = rc.password ?: "";
-		var postLoginUrl = Len( Trim( rc.postLoginUrl ?: "" ) ) ? rc.postLoginUrl : websiteLoginService.getPostLoginUrl( cgi.http_referer );
+		var postLoginUrl = websiteLoginService.getPostLoginUrl( explicitUrl=rc.postLoginUrl ?: "", defaultUrl=cgi.http_referer ?: "" );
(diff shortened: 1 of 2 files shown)


system/handlers/Login.cfc
```diff
@@ -107,7 +107,7 @@ component {
 		}
 
 		args.allowRememberMe = _getRememberMeAllowed();
-		args.postLoginUrl    = websiteLoginService.getPostLoginUrl( rc.postLoginUrl ?: event.getCurrentUrl() );
+		args.postLoginUrl    = !isEmpty( rc.postLoginUrl ?: "" ) ? rc.postLoginUrl : websiteLoginService.getPostLoginUrl( rc.postLoginUrl ?: event.getCurrentUrl() );
```
DominicWatson: I think this needs a little more work. Looking at the `getPostLoginUrl()` method shows that it sets this value into your session scope if its not already there and this will stop happening with this change. I suggest:

```
args.postLoginUrl = websiteLoginService.getPostLoginUrl(
      explicitUrl = rc.postLoginUrl ?: ""
    , defaultUrl  = event.getCurrentUrl()
);
```

Then change the logic in `websiteLoginService.getPostLoginUrl()` to always use the `explicitUrl` if it is not empty + to set it into session if using it. Otherwise, try using what is in session. Otherwise, use the defaultUrl and set it into the session.

DominicWatson (changes requested):
Hi Johnson, thanks for this. Please see review comment

system/handlers/Login.cfc
john5onCheng: Hi @DominicWatson  I have made a changes accordingly. It will now pickup explicitURL as priority, save it if available, otherwise look for the value from session or default value

PR #750, opened 2020-03-24
Feature presidecms 1801 prevent delete email blueprint

annbelievable:
Hi,

This is to prevent the admin from deleting the email blueprint if it is being used by email templates, please review.

Thank you,
Kok Ann


DominicWatson:
Nice one, thanks @annbelievable . FYI, I fixed up missing var scopes on the local variables in there before pushing. Will be in 10.11.39

PR #751, opened 2020-03-24
Feature presidecms 1670 form tab fieldset

annbelievable:
Hi,

This is to ensure all the forms' tabs and fieldsets in preside has id.

Thank you,
Kok Ann


DominicWatson:
Nice one, thanks @annbelievable - will be released into 10.11.39

PR #756, opened 2020-04-17
PRESIDECMS-1833 Mirror data export selectData arguments and field process…

vivienwong:
…ing with grid listing's setting


DominicWatson:
Nice one, thanks @vivienwong. Released in 10.11.43

PR #757, opened 2020-04-21
Update viewRecord.cfm

marcusegger:
if you have relations in your objects and therefore you have datamanager grid fields in your admin record view, you end up with an ugly view record layout just for the sake to have the system properties in the right col. 
The cols are 6/6 at the moment, the datamanager grid fields don't fit in 6 cold most of the time and so they overflow.
My proposal is to show the system information at the end of the admin record view and show the important information first.

### DIFF
--- a/system/views/admin/datahelpers/viewRecord.cfm
+++ b/system/views/admin/datahelpers/viewRecord.cfm
@@ -13,12 +13,12 @@
 	#preRenderRecord#
 
 	<div class="row">
-		<div class="col-md-6">
+		<div class="col-lg-12">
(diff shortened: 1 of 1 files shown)


DominicWatson:
Hey @marcusegger, thanks for the PR.

While I do not disagree with what is proposed here, I don't think we can just change it like this due to "breaking" existing views for other installations. I recommend that you create your own extension for a "better view record screen" and start with this approach by overriding the view.

We have done this internally to have a completely different view record layout that can include our own company norms, etc.

PR #760, opened 2020-04-24
Update cms_de.properties

marcusegger:
added explanation for user where to change filter later


DominicWatson:
Fab, thanks @marcusegger. Will be in 10.11.43

PR #761, opened 2020-04-24
Update cms.properties

marcusegger:
added explanation for user where to change filter later


DominicWatson:
Brill, thanks @marcusegger - this is great. Will be in 10.11.43

PR #765, opened 2020-05-07
PRESIDECMS-1794

john5onCheng:
Add options to show more meaningful message and allow old reset password token authenticated
Add options to restrict multiple reset password within an editable time frame.


DominicWatson:
Thanks for this extensive PR @john5onCheng - merging it into the upcoming 10.12.0 release now and we will do some testing and validation around the features internally. 

Looks great!

PR #766, opened 2020-05-14
PRESIDECMS-1849 trigger record activity only if send log is updated

jtcy0123:
To reduce error in creating send log activity where the send log does not exist


DominicWatson:
Brilliant. Thanks for the attention to detail @jtcy0123 :)

PR #767, opened 2020-05-16
PRESIDECMS-1852

john5onCheng:
Set to allow custom config from handler


system/assets/js/admin/presidecore/preside.richeditor.js
```diff
@@ -18,6 +18,7 @@ PresideRichEditor = ( function( $ ){
 		  , enterMode             = $elementToReplace.data( "enterMode" )
 		  , autoParagraph         = $elementToReplace.data( "autoParagraph" ) !== undefined ? $elementToReplace.data( "autoParagraph" ) : cfrequest.ckeditorAutoParagraph
 		  , defaultConfigs        = cfrequest.ckeditorDefaultConfigs || {}
+		  , customDefaultConfigs  = $elementToReplace.data( "custom-default-configs" ) || {}
```
sebduggan, line 21: Hi @john5onCheng 

Although `.data( "custom-default-configs" )` will work, the jQuery convention is that dashes in the HTML data attribute get converted to camelCase in jQuery.

So preferable would be `.data( "customDefaultConfigs" )`, similar to the lines above.

DominicWatson:
This has been merged into 10.12.0, thanks @john5onCheng.

PR #768, opened 2020-05-19
PRESIDECMS-1853: Check for etag header in http request data

smoufa:
The etag header sent by the client isn't recognised in method _doBrowserEtagLookup().

The committed code worked in Chrome and Firefox with Lucee 5.3 and CommandBox.

Caching time for the i18n file in _serveI18nBundle() is still set to 20 minutes but should be much smaller.

### DIFF
--- a/system/handlers/admin/StaticAssetDownload.cfc
+++ b/system/handlers/admin/StaticAssetDownload.cfc
@@ -19,7 +19,7 @@ component {
 		_doBrowserEtagLookup( etag );
 
 		header name="cache-control" value="max-age=31536000";
-		header name="etag" value=etag;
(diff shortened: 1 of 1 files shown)


DominicWatson:
Nice one, thanks @smoufa.

> Caching time for the i18n file in _serveI18nBundle() is still set to 20 minutes but should be much smaller.

I think really the answer to this one is to have some kind of system where we include the current lancuage in the URL to fetch the file. Then the cache time of 20m should be fine.

DominicWatson:
I have created a separate ticket for that:

https://presidecms.atlassian.net/browse/PRESIDECMS-1854

PR #774, opened 2020-06-04
Feature presidecms 1867 scheduled task to purge expired workflow

chunwengpixl8:
Added a scheduled task where it will purge the expired workflow data based on the 'settings.purgeWorkflowStateTimeFrame' configured under config file.


DominicWatson:
Nice one, thanks @chunwengpixl8 . I've merged into upcoming 10.12.0 release.

Note, I removed the time frame configuration option as I wasn't sure it was the right place or variable name. If we absolutely need this we can add it in later (I think a preside admin setting would be preferable).

PR #775, opened 2020-06-07
PRESIDECMS-1179

john5onCheng:
Allow site tree page to return result in order

### DIFF
--- a/system/services/siteTree/SiteTreeService.cfc
+++ b/system/services/siteTree/SiteTreeService.cfc
@@ -189,12 +189,15 @@ component {
 		, array   ids          = []
 		, array   extraFilters = []
 	) {
-		var filter = { "page.trashed" = false };
-		var extra  = Duplicate( arguments.extraFilters );
-		var params = {};
+		var filter  = { "page.trashed" = false };
+		var extra   = Duplicate( arguments.extraFilters );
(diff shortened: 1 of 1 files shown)


system/services/siteTree/SiteTreeService.cfc
```diff
@@ -189,12 +189,15 @@ component {
...
-		var params = {};
+		var filter  = { "page.trashed" = false };
+		var extra   = Duplicate( arguments.extraFilters );
+		var params  = {};
+		var orderBy = "page._hierarchy_sort_order";
 
 		if ( arguments.ids.len() ) {
 			extra.append( { filter={ "page.id"=arguments.ids } } );
+			params.ids = { list=true, type="varchar", value=arguments.ids }
+			orderBy = "FIELD( page.id, :ids )"
```
sebduggan, line 200: Please note that `FIELD()` is MySQL-specific and not supported by MSSQL, so this code will fail if run in MSSQL...

DominicWatson:
Indeed, given that, `Field()` is MySQL specific, I think it is not possible to merge at this point in time. We'll need to look further into why this would be happening. I think if it is about displaying already selected values, then the order of the ajax results is irrelevant and it will be down to whatever js is populating the uber picker.

PR #779, opened 2020-06-10
PRESIDECMS-1871 Switch to include blank option

vivienwong:
For the flexibility of not pre-selecting anything


sebduggan:
Thanks @vivienwong - merged with a small change: the argument is now `includeEmptyOption`, and will render placeholder text (if provided) as the option label.

PR #781, opened 2020-06-11
PRESIDECMS-1874 Additional condition check before including field to …

vivienwong:
…export field options


sebduggan:
Thanks @vivienwong - merged and released.

PR #783, opened 2020-06-18
Add show hide password option

john5onCheng:


system/views/formcontrols/password/index.cfm
```diff
@@ -15,13 +15,22 @@
...
+		event.include( "/css/admin/specific/password/" );
+	}
+
 	passwordPolicyContext = args.passwordPolicyContext ?: "";
 </cfscript>
 
 <cfoutput>
 	<input type="password" id="#inputId#" placeholder="#placeholder#" name="#inputName#" tabindex="#getNextTabIndex()#" value="#value#" class="#inputClass# form-control"<cfif Len( Trim( passwordPolicyContext ) )> data-password-policy-context="#passwordPolicyContext#"</cfif>>
+	<cfif IsTrue( args.allowShowHidePassword ?: "" )>
+		<a toggleid="###inputId#" class="fa fa-fw fa-eye toggle-password" title="Help" data-rel="popover" data-trigger="hover" data-placement="left" data-content="Show / hide the input value"></a>
```
DominicWatson: Few things on this line:

* `toggleid` should either just be `href="###inputId#"` or `data-target="###inputId#"` (and adjust the javascript to use that)
* `title="Help"` should use i18n
* `data-content` should use i18n

DominicWatson (changes requested):
Hi @john5onCheng this looks great thank you. Some feedback on the view but otherwise good.

system/views/formcontrols/password/index.cfm
john5onCheng: Hi Dom,

I have updated the code accordingly

Many thanks
J

PR #784, opened 2020-06-23
PRESIDECMS-1881 there is no property with name [1]  found in [java.util.LinkedHashSet]

guustnieuwenhuis:
Fixed issues related to getting back a LinkedHashSet rather than a String List


DominicWatson:
Brilliant, thanks for the extra work on this @guustnieuwenhuis. Available in:

10.12.0-SNAPSHOT6519

PR #788, opened 2020-07-08
PRESIDECMS-1893: set autocomplete="new-password" on new password fields

smoufa:


DominicWatson:
Nice work - thanks for this @smoufa - released in 10.11.59.

PR #789, opened 2020-07-09
PRESIDECMS-1895

john5onCheng:
Add new interception point for dummy page


DominicWatson:
Brill, thanks @john5onCheng - releasing in 10.11.59 hotfix.

PR #791, opened 2020-07-17
PRESIDECMS-1901 filter by field on object picker not working

jtcy0123:
Remove the additional s from DataManager.cfc$getObjectRecordsForAjaxSelectControl filterByField to match the one on query string from ObjectPicker


DominicWatson:
Hey thanks for this @jtcy0123. I've seen this working in many places - but perhaps this is something more specific that I am missing.

I think it would make sense to do:

```
var filterByFields = ListToArray( rc.filterByField ?: ( rc.filterByFields ?: "" ) );
```

Just in case anyone is relying on `filterByFields`. Could you make that change? I'll then release this as a hotfix.

jtcy0123:
Thanks @DominicWatson, have made the changes:
`var filterByFields = ListToArray( rc.filterByField ?: ( rc.filterByFields ?: "" ) );`

PR #799, opened 2020-07-24
Feature fix filter toggle

seakchiew:
small fix on the filter toggle on the data manager admin object listing page where instance id is missing from the target


DominicWatson:
Great, thanks @seakchiew - merging into 10.12.0 that will be released soon.

PR #800, opened 2020-07-27
PRESIDECMS-1903 control batch delete edit allowed operations

pixl8-brayden:
- add object attribute _batchEditable_ to have object level control, default to true
- add object attribute _batchDelete_ to have record delete but restricted batch delete, default to true

### DIFF
--- a/system/config/Config.cfc
+++ b/system/config/Config.cfc
@@ -285,7 +285,7 @@ component {
 			  cms                    = [ "access" ]
 			, sitetree               = [ "navigate", "read", "add", "edit", "activate", "publish", "savedraft", "trash", "viewtrash", "emptytrash", "restore", "delete", "manageContextPerms", "viewversions", "sort", "translate", "clearcaches", "clone" ]
 			, sites                  = [ "navigate", "manage", "translate" ]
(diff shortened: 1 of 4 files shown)


DominicWatson:
I think it makes more sense to add this to the list of operations. i.e. `isOperationAllowed( "batchedit" )` / `isOperationAllowed( "batchdelete" )`. 

This would default to `true` unless `@datamanagerAllowedOperations` did not include it or `@datamanagerDisallowedOperations` contained either `edit` _or_ `batchedit` (and `delete` _or_ `batchdelete` for the batchdelete operation).

This has some backward compatibility issues but I'm happy enough with that. Nothing major I'd suggest.

So `batchedit` is **not** allowed when:

* `edit` is a disallowed operation
* `batchedit` is a disallowed operation
* `batchedit` is not included in an explicit allowed operations list

`batchdelete` is **not** allowed when:

* `delete` is a disallowed operation
* `delete` is a disallowed operation
* `batchdelete` is not included in an explicit allowed operations list

pixl8-brayden:
Hi Dom,

I had updated the control to using data manger operations.

Thanks

pixl8-brayden:
Hi @DominicWatson 

The controls had amend to use data manager operations and added admin permission for data manager.

Thanks

DominicWatson:
Perfect, looks great - will merge into 10.12.0 and get some further QA on it.

PR #801, opened 2020-07-27
PRESIDECMS-1910 limit data manager sortable columns in grid fields

pixl8-brayden:
add _noSortableFields_ object attributes to control which fields are not sortable.

### DIFF
--- a/system/assets/js/admin/specific/datamanager/object/listingTable.js
+++ b/system/assets/js/admin/specific/datamanager/object/listingTable.js
@@ -124,6 +124,7 @@
 				datatable = $listingTable.dataTable( {
 					aoColumns     : colConfig,
 					aaSorting     : defaultSort,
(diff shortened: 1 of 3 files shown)


DominicWatson:
Neat solution, thanks @pixl8-brayden . However, `noSortableFields` is a strange expression I think.

I would prefer it the other way around. All fields are sortable unless you specify a list of sortable fields:

```
@datamanagerSortableFields title,datecreated,datemodified,blah
```

Then reverse all the logic you made there.

pixl8-brayden:
Hi @DominicWatson 

The object attribute is updated to _datamanagerSortableFields_ and will append the _no-sorting_ class only when object's datamanagerSortableFields is supplied and no contained.

Thanks

DominicWatson:
Perfect, thanks @pixl8-brayden - merged into 10.12.0 and building now.

PR #802, opened 2020-07-27
PRESIDECMS-1905 fix internal date fields in excel export do not use labels in column headers

pixl8-brayden:
add i18n translate for preside object default fields


DominicWatson:
Great, thanks @pixl8-brayden - merged into 10.12.0

PR #804, opened 2020-07-27
PRESIDECMS-1909 object picker form control default sort order

pixl8-brayden:
objectPicker form control use data grid default sort order if not specify in the form field.


DominicWatson:
brill, thanks @pixl8-brayden - merging into release 10.12.0

PR #805, opened 2020-07-27
PRESIDECMS-1906 possibility to define the default exporter

pixl8-brayden:
add _settings.dataExport.defaultExporter_ for application config.cfc defined default exporter.


DominicWatson:
Brill thanks @pixl8-brayden - merged into upcoming 10.12.0 release

PR #807, opened 2020-07-28
PRESIDECMS-1917 error setup password policy when use null value in detail policy setting

pixl8-brayden:
add val cast to make sure it always pass in numeric value.


DominicWatson:
Great, thanks @pixl8-brayden - have merged in with the PRESIDECMS-1913 work and into 10.12.0.

PR #808, opened 2020-07-29
PRESIDECMS-1912 switch action order in object history listing

pixl8-brayden:
- __version_is_draft_ is deprecated in 10.12, update to use __version_is_latest_ column instead in DataManagerService$getRecordHistoryForGridListing
- switch data manager history listing actions view and edit order to let view action 1st.


DominicWatson:
Nice one, merged into 10.12.0

PR #809, opened 2020-07-29
PRESIDECMS-1885 grid listing header label improvements

pixl8-brayden:
- add new proxies helper to get base uri root
- add new context argument for i18n$translatePropertyName
- add optional tooltip (help) for data table listing header
i18n file sample
data table listing sample

### DIFF
--- a/system/views/admin/datamanager/_objectDataTable.cfm
+++ b/system/views/admin/datamanager/_objectDataTable.cfm
@@ -171,7 +171,15 @@
 						</th>
 					</cfif>
 					<cfloop array="#args.gridFields#" index="fieldName">
-						<th data-field="#ListLast( fieldName, '.' )#">#translatePropertyName( args.objectName, fieldName )#</th>
+						<th data-field="#ListLast( fieldName, '.' )#">
+							#translatePropertyName( args.objectName, fieldName )#
+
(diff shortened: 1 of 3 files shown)


system/views/admin/datamanager/_objectDataTable.cfm
```diff
@@ -171,7 +171,15 @@
 						</th>
 					</cfif>
 					<cfloop array="#args.gridFields#" index="fieldName">
-						<th data-field="#ListLast( fieldName, '.' )#">#translatePropertyName( args.objectName, fieldName )#</th>
+						<th data-field="#ListLast( fieldName, '.' )#">
+							#translatePropertyName( args.objectName, fieldName )#
```
DominicWatson: Should this have a context of 'listing', i.e.

```translatePropertyName( args.objectName, fieldName, "listing" )```

DominicWatson (commented):
Looks good, I just have a query about the implementation in the listing view. Could you check that?

Also, could you find somewhere to add to the documentation? :) I think somewhere in the datamanager section would make sense.

system/views/admin/datamanager/_objectDataTable.cfm
pixl8-brayden: thanks for notice this, my bad was missed out when copying over.

DominicWatson:
Fab, thanks @pixl8-brayden - merging into 10.12.0

PR #810, opened 2020-07-29
PRESIDECMS-1908 ability to omit field labels for fields in single record view

pixl8-brayden:
add new _ displayPropertyTitle_ property attribute to control visibility of property title, default is true.


DominicWatson:
Nice one, thanks @pixl8-brayden  - could you also update the documentation in https://docs.preside.org/devguides/datamanager/adminrecordviews.html ?

Merging into 10.12.0.

PR #811, opened 2020-07-29
PRESIDECMS-1911 permissions on advanced filter builder usage

pixl8-brayden:
1. hide rule engine builder in object listing view if permission not granted, but still able to access favourite and saved filter.
Screenshot 2020-07-29 at 16 03 50

2. add new _rules engine manage_ role to grant all aspect of control for rules engine feature.


DominicWatson:
Looking good @pixl8-brayden - nice one. Will need to make some release notes around the permissions changes and new role.

DominicWatson:
Merging into 10.12.0

PR #812, opened 2020-08-06
Filter params for email mass send queue subquery

choontat-pixl8:
Ensure that params from email mass send queue subquery are included in the filter params for duplicate check filter


DominicWatson:
Perfect, thanks @choontat-pixl8 - going into 10.12.1 hotfix.

PR #813, opened 2020-08-06
PRESIDECMS-1873 data exists is not as efficient as it should be

pixl8-brayden:
use _recordCountOnly_ argument for data exist and check the return result is always in boolean format.


DominicWatson:
Nice one, thanks @pixl8-brayden - had thought this one had already been done! Going to be released in 10.12.1

PR #817, opened 2020-08-14
PRESIDECMS-1883 cloning of rules engine conditions and filters

pixl8-brayden:
- add ability to clone of existing rules engine conditions and filters.


DominicWatson:
Looking good :) merging into 10.13.0 upcoming release for testing.

PR #819, opened 2020-08-19
PRESIDECMS-1937 new left right multi select panel form control

pixl8-brayden:
Add new _MultiSelectPanel_ form control for left right multi select panel layout
Screenshot 2020-08-19 at 10 09 04
- Able to explicitly provide values and labels for select options.
- Able to provide object and default will get records (default fields `id, label`) as options
- If object is provided, able to use object properties as options by set `useObjProperties` to true
- Sortable default to false (sorting panel will be hidden if not sortable)

Add new _dataExportPanelPicker_ form control to replace existing data export field picker control (extended from _MultiSelectPanel_ with `useObjProperties=true`, `args.sortable =true` and get object from request context)
Screenshot 2020-08-19 at 11 00 42

### DIFF
--- a/system/assets/css/admin/specific/multiSelectPanel/multiSelectPanel.less
+++ b/system/assets/css/admin/specific/multiSelectPanel/multiSelectPanel.less
@@ -1,0 +1,23 @@
+@import url( 'globals.less' );
+
+.multi-select-panel {
+	display: flex;
+
+	.action-buttons {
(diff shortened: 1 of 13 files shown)


sebduggan:
Thanks @pixl8-brayden - this is merged into the 10.13.0 release.

DominicWatson:
Awesome :)

PR #820, opened 2020-09-01
PRESIDECMS-1939

john5onCheng:
Add option to specify site domain on build admin link


DominicWatson:
@john5onCheng is there any reason this should only be for admin links and not all other types of links? Can you give an example usage?

john5onCheng:
Ah, I got your point. That should apply to the other routehandler too. I create this PR when working on a multi site admin project which required to send welcome email which containing link to correct microsite.
e.g.
User A from Site A
User B from Site B

User B should link with Site B but not Site A which is default site.

PR #821, opened 2020-09-03
PRESIDECMS-1950 parsedatetime in getPrefetchCachebusterForAjaxSelect

flolopdel:
Fix issue when we are working with cbehcache after deserializing a date field from a query


DominicWatson:
Brill thanks @flolopdel - will be releasing into 10.12.5 hotfix after testing.

PR #829, opened 2020-09-15
PRESIDECMS-1956 Dialog confirmation prompt with matching text

teonator:
Update dialog with optional `data-confirmation-match="hello-world"` to enable matching text confirmation.


system/assets/js/admin/presidecore/preside.confirmation.prompts.js
```diff
@@ -17,15 +18,60 @@
...
+		var $message = $( "<div class=\"form-group\"><label>" + $link.data( "confirmationPrompt" ) + "</label></div>" )
+		  , $input   = $( "<input class=\"bootbox-input form-control\" autocomplete=\"off\" type=\"text\" />" )
+		  , match    = ""
+		;
+
+		if( typeof $link.data( "confirmation-match" ) !== "undefined" ) {
+			match =  $link.data( "confirmation-match" );
+
+			$message
+				.append( "<p class=\"help-block\">Please type <code>" + match + "</code> to proceed.</p>" )
```
DominicWatson: This should use the i18n system. can do this in js:

```js
i18n.translateResource( "cms:confirmation.prompt.please.type.message", [ "<code>" + match + "</code>" ] );
```

Then add this to `/i18n/cms.properties` file:

```properties
confirmation.prompt.please.type.message=Please type {1} in the input below to proceed.
```

DominicWatson (changes requested):
Love the simplicity of this. Could you check on the feedback re i18n though? Need to ensure that all admin text can be translated :+1: 

Otherwise looks great.

system/assets/js/admin/presidecore/preside.confirmation.prompts.js
teonator: I have updated the i18n for message. I also added i18n for confirm and cancel buttons.

DominicWatson:
L33t. Merging into upcoming 10.13.0 for internal QA.

teonator:
I have updated view record's top right button and listing's multi action for the confirmation prompt with matching text.

DominicWatson:
Great, I've merged these changes into the upcoming release. In future though, please avoid merging the release branch into your feature. Creates an unnecessarily large diff and makes the feature co-dependent on all the other release features.

DominicWatson:
This has actually been merged. Closing.

PR #830, opened 2020-09-22
PRESIDECMS-1957 fix many to many of draft objects not shown in record view

pixl8-brayden:
- update `_selectFromVersionTables` (return query) private function to `_prepareSelectFromVersionTables` (return prepared struct) and run sql in selectData function instead.
- update `relatedRecordsDatatable` function to check use versioning and record has draft to query from version table.

### DIFF
--- a/system/handlers/admin/DataHelpers.cfc
+++ b/system/handlers/admin/DataHelpers.cfc
@@ -163,10 +163,10 @@ component extends="preside.system.base.adminHandler" {
 	 *
 	 */
 	public void function getRecordsForRelatedRecordsDatatable( event, rc, prc ) {
-		var objectName       = rc.objectName       ?: "";
-		var propertyName     = rc.propertyName     ?: "";
-		var recordId         = rc.recordId         ?: "";
-		var fromVersionTable = rc.fromVersionTable ?: false;
+		var objectName       = rc.objectName   ?: "";
+		var propertyName     = rc.propertyName ?: "";
+		var recordId         = rc.recordId     ?: "";
+		var fromVersionTable = isTrue( rc.fromVersionTable ?: "" );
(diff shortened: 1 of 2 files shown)


system/handlers/admin/DataHelpers.cfc
```diff
@@ -161,18 +163,20 @@ component extends="preside.system.base.adminHandler" {
...
-		var recordId       = rc.recordId     ?: "";
-		var gridFields     = adminDataViewsService.listGridFieldsForRelationshipPropertyTable( objectName, propertyName ).toList();
-		var relatedObject  = presideObjectService.getObjectPropertyAttribute( objectName=objectName, propertyName=propertyName, attributeName="relatedTo" );
-		var relatedIdField = presideObjectService.getIdField( objectName=relatedObject );
-		var extraFilters   = [];
-		var subquerySelect = presideObjectService.selectData(
+		var objectName       = rc.objectName       ?: "";
+		var propertyName     = rc.propertyName     ?: "";
+		var recordId         = rc.recordId         ?: "";
+		var fromVersionTable = rc.fromVersionTable ?: false;
```
DominicWatson: You can ensure that this variable is a boolean (so no need for BooleanFormat(), below), with:

```
var fromVersionTable = isTrue( rc.fromVersionTable ?: "" );
```

That is the preferred approach here.

DominicWatson (changes requested):
Hey this looks pretty perfect. I've made some tiny feedback that would be good to pick up, but otherwise fantastic stuff - really appreciated.

system/handlers/admin/DataHelpers.cfc
pixl8-brayden: this is updated.

DominicWatson:
Perfect, thanks @pixl8-brayden - merging into upcoming release for QA before we release hotfix.

PR #831, opened 2020-09-29
Make formula field to bypass default numeric renderer.

karhong:
Update the default numeric renderers to skip formula field, to prevent formula field always rendered as float.


DominicWatson:
Nice one, thanks @karhong - that's building into upcoming 10.13 snapshot now.

PR #834, opened 2020-10-02
PRESIDECMS-1966 restrict operations for used questions

pixl8-brayden:
Delete button remove and warning alert when question is in used.
Screenshot 2020-10-02 at 15 53 33

Denied multi delete actions if one of selected questions is in used.
Screenshot 2020-10-02 at 15 53 49

### DIFF
--- a/system/handlers/admin/datamanager/formbuilder_question.cfc
+++ b/system/handlers/admin/datamanager/formbuilder_question.cfc
@@ -232,9 +232,9 @@ component extends="preside.system.base.AdminHandler" {
 
 			for ( var action in args.actions ) {
 				if ( find( "deleteRecordAction", action.link ?: "" ) ) {
-					action.link  = "##";
-					action.class = "disabled";
 					action.title = "";
+					action.class = "disabled";
+					action.link  = "##";
 					break;
 				}
 			}
--- a/system/handlers/admin/datamanager/formbuilder_question.cfc
+++ b/system/handlers/admin/datamanager/formbuilder_question.cfc
@@ -2,6 +2,7 @@ component extends="preside.system.base.AdminHandler" {
 
 	property name="datamanagerService"          inject="datamanagerService";
 	property name="formBuilderItemTypesService" inject="formBuilderItemTypesService";
+	property name="formBuilderQuestionService"  inject="formBuilderQuestionService";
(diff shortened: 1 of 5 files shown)


system/handlers/admin/datamanager/formbuilder_question.cfc
```diff
@@ -188,6 +199,45 @@ component extends="preside.system.base.AdminHandler" {
...
+		var record     = args.record     ?: {};
+		var recordId   = record.id       ?: "";
+		var isInUse    = formBuilderQuestionService.questionIsInUse( questionId=recordId );
+
+		if ( isTrue( isInUse ) ) {
+			args.actions = args.actions ?: [];
+
+			for ( var action in args.actions ) {
+				if ( find( "deleteRecordAction", action.link ?: "" ) ) {
+					arrayDelete( args.actions, action );
```
DominicWatson: Is there a way we can *disable* this instead of deleting it? Removing it messes with alignment in the grid. Perhaps: 

`.link = "##", class="disabled" ` (or similar)

DominicWatson (commented):
This looks perfect, thanks @pixl8-brayden - just a little UX thing re the grid icons that would be great to fix before merging in.

system/handlers/admin/datamanager/formbuilder_question.cfc
```diff
@@ -188,6 +199,48 @@ component extends="preside.system.base.AdminHandler" {
...
+		var record     = args.record     ?: {};
+		var recordId   = record.id       ?: "";
+		var isInUse    = formBuilderQuestionService.questionIsInUse( questionId=recordId );
+
+		if ( isTrue( isInUse ) ) {
+			args.actions = args.actions ?: [];
+
+			for ( var action in args.actions ) {
+				if ( find( "deleteRecordAction", action.link ?: "" ) ) {
+					action.link  = "##";
```
sebduggan, line 235: Might be better to use `link = "javascript:void(0)"` here.

As it stands, the link is still active and will alter the URL. Not a big difference, but ensures the link does absolutely nothing.

(Tip: you can also just say `"javascript:"` and it will have the same effect).

sebduggan: Change made when merging PR

PR #835, opened 2020-10-02
PRESIDECMS-1958 Add custom questions view forms.

karhong:
Add custom questions view forms


DominicWatson:
Nice one @karhong . I've simplified the logic somewhat. That said, what you've done here is really sound and I've kept the crux of what you've done and just removed the unnecessary logic. The changes I made are here:

https://github.com/pixl8/Preside-CMS/commit/a87eb16c91416fcbc45c047c85e98f1d4cbf26be

PR #836, opened 2020-10-02
Add alias to pseudo-column in dataExists method

jjannek:
using "1" in the sql select list breaks the query for recordcount-only query in MSSQL. Using an Alias for the select field solves it.


DominicWatson:
Nice one, will get this into 10.13 to verify before including in hotfix. Thanks @jjannek

PR #838, opened 2020-10-05
Feature presidecms 1968 formbuilder widget being cached

mookeshNanda:
Formbuilder widget is being cached by the page causing it to persist initial formSubmission data.


DominicWatson:
I think the answer to this problem is slightly different. The answer is _not_ to disable the page cache, the form itself is not cached. Instead, the data that was submitted is what is being cached (we cache both page content and also certain rc and prc data). So the answer should be to ensure any submitted form data is never cached.

PR #841, opened 2020-10-08
remove the variable that does not exist

annbelievable:
there is some code that doesnt belong to UserLastVisitedPage.cfc's evaluateExpression method


DominicWatson:
Thanks for this @annbelievable - could you target *stable* however, and hot a specific hotfix? Am closing this, but could you reopen targeting the correct branch?

TIA

PR #844, opened 2020-10-14
PRESIDECMS-1958 form i18n typo

teonator:
Fix i18n formbuilder item type typo.


DominicWatson:
Nice one, will merge into the formbuilder branch - if folks have lived with these typos for this long they can wait for the release ;)

PR #845, opened 2020-10-14
PRESIDECMS-1970_FormBuilder_Questions_Rules_Engine

Pixl8RusselCole:


system/handlers/rules/expressions/question/UserHasResponse.cfc
```diff
@@ -0,0 +1,47 @@
...
+	 */
+	private boolean function evaluateExpression(
+		  required string question
+		,          boolean _has = true
+	) {
+		var filter = prepareFilters( argumentCollection = arguments	) ;
+
+		return formBuilderFilterService.evaluateQuestionUserHasResponse(
+			  argumentCollection = arguments
+			, userId             = payload.user.id
```
DominicWatson: I think we need to defend against this variable not existing. i.e. if there is no logged in user and this condition is run (in which case we can presumably immediately return:

```return !arguments._has;```

This would obviously need to be done for all the user based _expressions_ (the filters remain unchanged).

File: system/handlers/rules/fieldtypes/FormbuilderQuestionSingle.cfc [hunk 2]
```diff
@@ -0,0 +1,60 @@
...
+		var values = [];
+		var labels = [];
+		var objectUriRoot = presideObjectService.getResourceBundleUriRoot( "formbuilder_question" );
+		var label  = translateResource( objectUriRoot & "title" )
+
+		for ( var question in singleQuestions ) {
+			arrayAppend( values, question.id );
+			arrayAppend( labels, question.field_label );
+		}
+			if ( values.len() ) {
```
DominicWatson: Double indented (whitespace)

system/services/formbuilder/FormBuilderFilterService.cfc
```diff
@@ -0,0 +1,1095 @@
...
+		if ( !$isWebsiteUserLoggedIn() ) {
+			return false;
+		}
+
+		if ( !len( trim( arguments.conditionId ) ) ) {
+			return false;
+		}
+
+
+		var match = _getRulesEngineConditionService().evaluateCondition(
```
DominicWatson: Stylisticly here, I think we can just return the result of the function call, rather than storing in a variable. i.e.

```
// this
return someFunctionCall( ... );

// not this
var result = someFunctionCall( ... );
return result;
```

Unless we do something with result before returning it.

system/services/formbuilder/FormBuilderFilterService.cfc
```diff
@@ -0,0 +1,1095 @@
...
+		, required string value
+		,          boolean _all              = false
+		,          string parentPropertyName = ""
+		,          string filterPrefix       = ""
+	) {
+		var paramSuffix         = _getRandomFilterParamSuffix();
+		var responseQueryAlias  = "responseCount" & paramSuffix;
+		var overallFilter       = "#responseQueryAlias#.response_count >= 1 ";
+		var params              = {
+								  "question#paramSuffix#" = { value=arguments.question, type="cf_sql_varchar" }
```
DominicWatson: whitespace (weird indentation)

DominicWatson (changes requested):
Russ, this looks amazing. Fab effort.

I've put some minor review comments in. Please check other occurrences (things like whitespace).

But once done can get this merged and into testing :)

system/services/formbuilder/FormBuilderFilterService.cfc
Pixl8RusselCole: I'd removed a writeDump(match) and should have changed it back to returning the result immediately

DominicWatson:
Awesome sauce, merging into the upcoming release for QA.

PR #846, opened 2020-10-15
PRESIDECMS-1958 strip html tags from anchor title attribute

teonator:
Created new stripTags() helper. Updated delete button to strip tags from label before assign to anchor title.

### DIFF
--- a/system/handlers/admin/DataManager.cfc
+++ b/system/handlers/admin/DataManager.cfc
@@ -1533,7 +1533,7 @@ component extends="preside.system.base.AdminHandler" {
 				, iconClass = "fa-trash"
 				, globalKey = "d"
 				, title     = translateResource( uri="cms:datamanager.deleteRecord.btn" )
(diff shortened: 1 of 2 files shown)


DominicWatson:
Question: should the behaviour not be that if I want HTML tags in the prompt then I _can_ - the HTML output should be escaped and then re-encoded when outputting the prompt?

Do we have an example where this has happened?

```cfm
<a ... data-title="#HtmlEditFormat( prompt )#">...</a>
```

teonator:
This is on data manager view record. The delete button shows the HTML tags in the tooltip when user hover on it. 
Using htmlEditFormat() will shows encoded HTML to user.

```
<a ... title="Delete the custom field, Hello World (&lt;code&gt;hello_world&lt;/code&gt;)?">...</a>
```

DominicWatson:
kk fair enough - will merge into the FB2 branch pending 10.13.0 release before this gets closed down.

PR #848, opened 2020-10-15
PRESIDECMS-1977

john5onCheng:
Allow OR operator for extraFilter merge


DominicWatson:
I'm not sure about this one. This could break certain features. i.e. where you have interceptors adding additional filters they will need to know that it is an AND, not an OR.

I think you should be able to achieve this OR list of filters in some other way?

PR #853, opened 2020-10-19
PRESIDECMS-1958 Update radio response to label instead of value.

teonator:


DominicWatson:
Brill, thanks @teonator

PR #858, opened 2020-10-20
PRESIDECMS-1985 Add webp to core asset manager file types

teonator:
Add webp i18n. Update file type picker to handle unknown file type i18n instead of showing NOT FOUND.

### DIFF
--- a/system/handlers/formcontrols/FileTypePicker.cfc
+++ b/system/handlers/formcontrols/FileTypePicker.cfc
@@ -4,28 +4,33 @@ component {
 
 	public string function index( event, rc, prc, args={} ) output=false {
 		var types = [];
+
 		args.values = [];
 		args.labels = [];
 
 		for( var assetTypeGroup in assetTypes ){
+			var groupType = translateResource( "filetypes:#assetTypeGroup#.group.type", translateResource( "filetypes:unknown.group.type" ) );
+
 			types.append( {
 				  value = assetTypeGroup
-				, label = translateResource( "filetypes:#assetTypeGroup#.picker.label" )
+				, label = translateResource( "filetypes:#assetTypeGroup#.picker.label", translateResource( "filetypes:unknown.picker.label" ) )
 			} );
 
 			for( var type in assetTypes[ assetTypeGroup ] ){
+				var default = translateResource( uri="filetypes:default.picker.label", data=[ groupType, type ]  );
+
 				types.append( {
 					  value = type
-					, label = translateResource( "filetypes:#type#.picker.label", translateResource( "filetypes:#assetTypeGroup#.picker.text", "Unknown" ) & ": #type#" )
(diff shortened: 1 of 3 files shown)


system/handlers/formcontrols/FileTypePicker.cfc
```diff
@@ -16,7 +16,7 @@ component {
 			for( var type in assetTypes[ assetTypeGroup ] ){
 				types.append( {
 					  value = type
-					, label = translateResource( "filetypes:#type#.picker.label" )
+					, label = translateResource( "filetypes:#type#.picker.label", translateResource( "filetypes:#assetTypeGroup#.picker.text", "Unknown" ) & ": #type#" )
```
DominicWatson: If we are going to do defaults here, we need to default everything. So two extra i18n URIs: 

```properties
unknown.group.type=Unknown
default.type.label={1}: {2}
```

Then, here:

```cfc
groupType = translateResource( "filetypes:#assetTypeGroup#.picker.text", translateResource( "filetypes:unknown.group.type" ) );
default = translateResource( uri="default.type.label", data=[ groupType, type ]  );
// ...
label = translateResource( "filetypes:#type#.picker.label", default )
```

system/i18n/filetypes.properties
```diff
@@ -36,7 +38,9 @@ ppam.picker.label=Document: ppam
 pptm.picker.label=Document: pptm
 potm.picker.label=Document: potm
 ppsm.picker.label=Document: ppsm
+webp.picker.label=Document: webp
```
DominicWatson: Should `webp` not be video?

Also, this should be added to `/system/config/Config.cfc$_getConfiguredFileTypes()`.

DominicWatson (commented):
Hi @teonator - thanks for this, looking good. Could you check the review items and I'll merge when done.

system/handlers/formcontrols/FileTypePicker.cfc
teonator: Updated with additional suggested i18n and related changes.

system/i18n/filetypes.properties
teonator: `webp` is an image. https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types.
Updated i18n and added it into config.

teonator: I also found and fixed typo `serveAsAttachmemnt` of svg mime type in config.

DominicWatson:
Perfect, looking good. Thanks for the typo fix too ;)

DominicWatson:
I created a new separate ticket for this (as not really related to form builder): https://presidecms.atlassian.net/browse/PRESIDECMS-1985

I've merged this into 10.13.0 upcoming release ready for internal QA.

PR #860, opened 2020-10-21
PRESIDE-1958 Formbuilder form group error highlight/ unhighlight.

teonator:


DominicWatson:
Great, thanks @teonator - nice solution. Will merge now.

PR #866, opened 2020-10-26
PRESIDECMS-1958 Range validator with comma

teonator:
Update range validator message format with comma.
Update range validator to handle comma number.


DominicWatson:
Nice one, thanks @teonator

PR #867, opened 2020-10-26
PRESIDECMS-1992 Lucee Spreadsheet upgrade to latest release

vivienwong:
Upgrade to latest release version and have tested with:
1. Datamanager export
2. Formbuilder export

Special note: There is an existing error prior to Lucee Spreadsheet upgrade on Datamanager object export > ExcelMultiSheets - left as it is with the upgrade

### DIFF
--- a/box.json
+++ b/box.json
@@ -27,7 +27,7 @@
     ],
     "dependencies":{
         "sticker":"1.2.2",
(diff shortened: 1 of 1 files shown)


DominicWatson:
Brilliant :) Thanks @vivienwong . Merging into upcoming 10.13.0 release for internal testing.

PR #868, opened 2020-10-26
PRESIDECMS-1958 Datepicker & Fileupload JS 

teonator:
- Change datepicker `event.include()` to view to prevent extension potential duplicates js conflict.
- Replace non-existence `filetype` validator in jquery.validate.js with `extension` validator in additional-methods.js.


DominicWatson:
Nice one, thanks @teonator

PR #870, opened 2020-10-27
PRESIDECMS-1987 Email templates picker without draft

teonator:


DominicWatson:
Nice one, thanks @teonator . Will merge into upcoming 10.13.0 release.

Am tweaking just one thing though - for backwards compatibility, the default behaviour of the function should not change so allowDrafts should be true. Then I'll tweak the picker control to not allow drafts.

(just because custom code may be using this service)

PR #873, opened 2020-10-28
PRESIDECMS-1958 Max min validators comma.

teonator:
Updated additional max and min validators to handle comma.


DominicWatson:
Looks good to me. Will merge in with form builder v2 work and get into the release.

PR #876, opened 2020-10-28
PRESIDECMS-1958 Fix formbuilder frontend datepicker js css 

teonator:


system/assets/js/frontend/formbuilder/datePicker/bootstrap.datepicker.js
```diff
@@ -1038,9 +1038,9 @@
 		},
 		headTemplate: '<thead>'+
 							'<tr>'+
-								'<th class="prev"><i class="fa fa-arrow-left"/></th>'+
+								'<th class="prev"><i class="font-icon font-icon-left"/>&lsaquo;</th>'+
```
DominicWatson: Do these classes work in raw Preside? I guess as this is the frontend it may not matter. But this could break existing implementations if they are already built to work with these classes?

DominicWatson (changes requested):
This looks great. Just the query around the css classes + also, could there be a more descriptive name than `script.js` for the custom picker logic? Perhaps `preside.frontend.datepicker.js` - so:

```
datepicker/
  bootstrap.datepicker.js
  preside.frontend.datepicker.js
```

system/assets/js/frontend/formbuilder/datePicker/bootstrap.datepicker.js
teonator: Currently there is no working CSS classes for both `font-icon-*` (_core) or `fa-*` in basic Preside. Agree we stick with modified original `fa-*` to prevent breaking classes.

PR #883, opened 2020-11-02
PRESIDECMS-1927: Rule expressions - User last submitted form builder form - variable action does not exist

khanh-nguyen-pixl8:
1. Remove checking "action" in some rules - expressions not needed.
2. Correct Rules - Expressions description


DominicWatson:
Great, thanks for this @khanh-nguyen-pixl8 - putting into a hotfix.

PR #885, opened 2020-11-03
PRESIDECMS-1958 Question response fileupload download

teonator:
Fixed broken download link for file upload on question response.


DominicWatson:
Nice one, thanks @teonator

PR #887, opened 2020-11-03
PRESIDECMS-1958 Fix formbuilder selectbox unable to render datamanagerObject response correctly

karhong:
Fix select control unable to render datamanagerObject value correctly.


DominicWatson:
Nice one, merging now. Thanks @karhong

PR #888, opened 2020-11-04
PRESIDECMS-1958 Number thousand decimal format

teonator:
Create new format function helper to format number with thousand separator and dynamic decimal points.

### DIFF
--- a/system/helpers/numberUtils.cfm
+++ b/system/helpers/numberUtils.cfm
@@ -1,29 +1,10 @@
-<cffunction name="numberThousandDecimalFormat" access="public" returntype="string" output="false">
+<cffunction name="presideStandardNumberFormat" access="public" returntype="string" output="false">
 	<cfargument name="number" type="numeric" required="true" />
-	<cfargument name="decimals" type="numeric" required="false" default="0" />
 
 	<cfscript>
-		var decimalPlaceholder = "";
-
-		if( round( arguments.number ) != arguments.number || arguments.decimals > 0 ) {
-			var decimalsCount = 0;
-
-			if( arguments.decimals > 0 ) {
-				decimalsCount = arguments.decimals;
-			}
-			else {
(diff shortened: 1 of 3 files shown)


system/helpers/numberUtils.cfm
```diff
@@ -0,0 +1,29 @@
+<cffunction name="numberThousandDecimalFormat" access="public" returntype="string" output="false">
+	<cfargument name="number" type="numeric" required="true" />
+	<cfargument name="decimals" type="numeric" required="false" default="0" />
+
+	<cfscript>
+		var decimalPlaceholder = "";
+
+		if( round( arguments.number ) != arguments.number || arguments.decimals > 0 ) {
+			var decimalsCount = 0;
```
DominicWatson: I think this could be simpler:

```cfc
var decimalPlaces = Len( ListRest( arguments.number, "." ) );
var numberFormat = decimalPlaces ? ",.#repeatString( "_", decimalPlaces )#" : ",";

return LsNumberFormat( arguments.number, numberFormat );
```

system/helpers/numberUtils.cfm
```diff
@@ -0,0 +1,29 @@
+<cffunction name="numberThousandDecimalFormat" access="public" returntype="string" output="false">
```
DominicWatson: The thousand is misleading in the name here - could be any precision. I think name this `presideStandardNumberFormat()` - we can then make it do more later if wanted.

DominicWatson (changes requested):
Hi @teonator - please see review comments. I think this could be simpler (but needs testing of course).

system/helpers/numberUtils.cfm
teonator: Updated for simpler code.

system/helpers/numberUtils.cfm
teonator: Rename helper name.

DominicWatson:
Perfect, thanks. Merging.

PR #894, opened 2020-11-09
PRESIDECMS-2005 Fix to maintain everyone group when it's required

vivienwong:
PRESIDECMS-2005 Fix to maintain everyone group when it's required


DominicWatson:
Thanks for this @vivienwong - I've taken the approach and changed it slightly. I've had to do this manually due to the branch that the bug was caused in still being active. Will be merged soon.

Thanks for the contribution.

PR #895, opened 2020-11-11
PRESIDECMS-1958 Question submitted by renderer

teonator:
Create new custom renderer for question submitted by to handle anonymous, website user and security user.


DominicWatson:
Looking good, thanks @teonator

PR #896, opened 2020-11-11
PRESIDECMS-1978 tweak filter scope should not disabled default value option

karhong:


DominicWatson:
:+1: nice one, merging now.

PR #900, opened 2020-11-12
Presidecms 2004 dbsync per object

cfmitrah:


system/services/presideObjects/SqlSchemaSynchronizer.cfc
```diff
@@ -56,6 +56,11 @@ component {
 				tableVersionExists = StructKeyExists( versions, "table" ) and StructKeyExists( versions.table, obj.meta.tableName );
 				tableExists        = tableVersionExists or _getTableInfo( tableName=obj.meta.tableName, dsn=obj.meta.dsn ).recordCount;
 
+				if( StructKeyExists( obj.meta, "dbsync" ) and not obj.meta.dbsync ) {
```
DominicWatson: Would prefer:

```cfc
if ( IsBoolean( obj.meta.dbsync ?: "" ) ) && !obj.meta.dbsync ) {
}
```

Given this is used throughout, could even add a helper function to make more readable:

```cfc
if ( _skipSync( obj.meta ) ) { // ...
```
```cfc
private boolean function _skipSync( required struct meta ) {
    return IsBoolean( arguments.meta.dbsync ?: "" ) && !arguments.meta.dbsync;
}
```

system/services/presideObjects/SqlSchemaSynchronizer.cfc
```diff
@@ -156,6 +161,10 @@ component {
 		,          struct relationships = {}
 
 	) {
+		var versions       = _getVersionsOfDatabaseObjects( [arguments.dsn] );
```
DominicWatson: It would be great if we didn't have to get all this data from the DB just in case we're going to _skip_ a column.

Is there a way where we could refactor this so that for fields doing `dbsync=false` that we don't need to have all the definition SQLs for the column?

system/services/presideObjects/SqlSchemaSynchronizer.cfc
```diff
@@ -190,17 +201,49 @@ component {
...
+				}
+			} else {
+				columnSql &= delim & column.definitionSql;
+			}
+			column.version = Hash( column.definitionSql );
 			delim = ", ";
 		}
 
-
+		colMeta = "";
```
DominicWatson, line 212: I _think_ this variable is not `var`'d?

DominicWatson (changes requested):
Hey @cfmitrah - this is great :). I have some initial feedback before merging into upcoming 10.13.0 release. Could you take a look?

Really appreciate the effort on this one.

system/services/presideObjects/SqlSchemaSynchronizer.cfc
cfmitrah: Hi @DominicWatson ,

I've done this changes please check latest [commit](https://github.com/pixl8/Preside-CMS/pull/900/commits/1a93a4da1c22efcbdb8ffddb69297e14478effc9)

system/services/presideObjects/SqlSchemaSynchronizer.cfc
cfmitrah: Hi @DominicWatson ,

I've done this changes please check latest [commit](https://github.com/pixl8/Preside-CMS/pull/900/commits/1a93a4da1c22efcbdb8ffddb69297e14478effc9)

system/services/presideObjects/SqlSchemaSynchronizer.cfc
cfmitrah: Hi @DominicWatson ,

I've already declared this to var scope in line number 171 function name : _generateTableAndColumnSql() , please check

system/services/presideObjects/SqlSchemaSynchronizer.cfc
```diff
@@ -190,21 +198,53 @@ component {
 			column.definitionSql = adapter.getColumnDefinitionSql( argumentCollection = args );
 			column.alterSql      = adapter.getAlterColumnSql( argumentCollection = args );
 			column.addSql        = adapter.getAddColumnSql( argumentCollection = args );
-			column.version       = Hash( column.definitionSql );
-
-			columnSql &= delim & column.definitionSql;
+			if ( _skipSync( colMeta ) ) {
+				dbColumn = QueryFilter( columnsFromDb, function( col ) {
```
DominicWatson: I still feel uneasy about this block of code. If we are skipping sync of a column, I'd like to see that we don't gather any information about it at all. This feels like it could be subject to failure because we still then check this column later on. 

Can this be reworked so that we explicitly _skip_ the column when looping the columns in the update logic, rather than setting the SQL to what we think it currently is?

DominicWatson (changes requested):

system/services/presideObjects/SqlSchemaSynchronizer.cfc
cfmitrah: @DominicWatson Let me check further & try to make the changes as per your request.

cfmitrah: @DominicWatson  I did the below functionality ( preside\system\services\presideObjects\PresideObjectReader.cfc ) to avoid property ( if dbsync=false ) 

image

but preside consider as that column as deprecated and renamed with __depricated__ in DB.

Is it okay for you ?

**NOTE: **  To avoid these problem previously I've implemented the functionality to get old version of schema . 

do you have any idea related to that.

DominicWatson: So, right now the logic in `_updateDbTable()` is:

```cfc
for( column in columnsFromDb ){
// ...
    if ( StructKeyExists( colsSql, columnName ) ) {
       // ... update column if necessary
    } else if ( !column.column_name contains "__deprecated__" ) {
       // ... deprecate the column
    }
}
```

In here, we need to add a check to see if there is a local matching property for the DB column that has `dbsync=false`. I think the cleanest place for this would be at the start of the loop. Something like:

```
for( column in dbColumns ) {
    if ( StructKeyExists( arguments.objectProperties, column.column_name ) && _skipSync( arguments.objectProperties[ column.column_name ] ) ) {
        continue;
    }
    // ...

}
```

?

cfmitrah: Thanks @DominicWatson  will check

cfmitrah: Hi @DominicWatson  , 

I've changed code as per you mentioned , please check. 

shall we change this feature in docs ? if yes means what are the file I need to look ?

guustnieuwenhuis:
We've created a set of tests to be able to validate the changes: https://bitbucket.org/weareorange/dbsync-test

The repository contains a stable version of Preside that has the changes of this PR applied. The README.md file explains the different tests that are set up and available to play with.

DominicWatson:
Briliant, thanks @guustnieuwenhuis. This has been merged into upcoming 10.14.0 release branch (can install either by checking the latest release snapshot in forgebox, or just: `box install presidecms@be`).

PR #902, opened 2020-11-24
PRESIDECMS-2009 Rules Engine Auto filter label updates

jjannek:
Using i18n property labels instead of target object labels for many-to-many and one-to-many auto filter expressions.
This is useful for all objects which have multiple relations to the same object or a relation to itself.


DominicWatson:
Nice one thanks @jjannek , merged into upcoming 10.13.0 release. I created this ticket to track:

https://presidecms.atlassian.net/browse/PRESIDECMS-2009

PR #904, opened 2020-12-02
PRESIDECMS-2014: Support higher ASCII and HTML encoded entities also when click tracking

jjannek:
The styleinliner already supported the feature emailStyleInlinerAscii - but the click tracking not.
In addition the click tracking used to use there jsoup doc.toString() method which eliminates html encoded entities.

The result is that emails with click tracking enabled look and behave differently if either higher ascii chars or html encoded entities are used.

The code changes proposed here will solve this and make sure that the general email content is the same, both with and without click tracking enabled.

### DIFF
--- a/system/services/email/EmailLoggingService.cfc
+++ b/system/services/email/EmailLoggingService.cfc
@@ -523,7 +523,11 @@ component {
 			}
 		}
 
-		return doc.toString();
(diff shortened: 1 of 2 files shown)


DominicWatson:
Fantastic, thanks @jjannek . Merging into 10.12.19 hotfix.

PR #906, opened 2020-12-08
PRESIDECMS-2020 fix saving condition saved as filter bug

pixl8-brayden:
remove `filter_object` from formData if is not convert to filter when editing rule condition.


DominicWatson:
Nice one, will merge into our upcoming release for testing before releasing as hotfix.

PR #908, opened 2020-12-10
Updated german translations

berndheitmann:
Better translation for sitetree.navigate. "\u00FCbersetzen" on english is translate.


smoufa:
That makes sense.

DominicWatson:
Hi @berndheitmann - thanks for this :)

@smoufa we need to get these into Zanata or they'll be lost when we pull translations down (translations must always go into Zanata as source).

Would you be interesting in joining the translation team @berndheitmann (no commitment required). You would have access to the translation database on our Zanata installation.

DominicWatson:
Thanks for this, this has now been manually merged in using Zanata.

PR #909, opened 2020-12-10
WIP PRESIDECMS-2022 search and filter using formula and enum fields on data objects

pixl8-brayden:


pixl8-brayden:
Close due to duplicated with another PR.

PR #915, opened 2020-12-30
PRESIDECMS-2037 FormBuilderFormPicker

teonator:


system/handlers/formcontrols/FormBuilderFormPicker.cfc
```diff
@@ -0,0 +1,17 @@
+component {
+
+	public string function index( event, rc, prc, args={} ) {
+		arguments.args.labels = [ "" ];
+		arguments.args.values = [ "" ];
+
+		var forms = getPresideObject('formbuilder_form').selectData( filter={ active=1 } );
```
DominicWatson, line 7: Prefer `active=true`. 

I'll merge this anyway and just updated it. But future note, if the field is boolean, we should use a strict boolean value rather than relying on knowing specific DB implementation is actually a bit field.

DominicWatson:
Nice one, thanks @teonator will add into 10.14

system/handlers/formcontrols/FormBuilderFormPicker.cfc
teonator: Noted. Thanks.

PR #916, opened 2021-01-02
PRESIDECMS-2036 Logger helper

teonator:


DominicWatson:
Looks good. I'd prefer the naming and argument order this way:

```cfc
logMessage( logger, severity, message );
```

Am going to just do that if that's ok. Also, if the logger arg may be null, I think we _may_ need to call it like this?

```cfc
function someFunction( any logger ) {
    logMessage( arguments.logger ?: NullValue(), "info", "My info message" );
}
```

Or does it work without the `?: NullValue()`?

teonator:
Sure. I am fine with that.

DominicWatson:
I have done the above and merged into 10.14.0

PR #917, opened 2021-01-04
PRESIDECMS-1929 track user downloads of documents have public urls

pixl8-brayden:


DominicWatson:
This looks great, but doesn't go quite far enough. We still want S3 (or whatever static storage) to do the heavy lifting of serving the asset content itself. Otherwise, we rely on Lucee/Java serving the file in a request which is a waste of memory (it is not nearly as efficient as serving the file from NGiNX/S3/whatever).

The additional work should be to 302 temporary redirect to the public URL of the asset if that URL is different from `/asset/{id}../`. This redirect would happen after any tracking logic and predownload interceptors but before the logic to serve the file.

This would be in `/preside/system/handlers/core/AssetDownload.cfc`.

system/services/routeHandlers/AssetRouteHandler.cfc
```diff
@@ -43,24 +43,49 @@ component implements="iRouteHandler" singleton=true {
...
+		var trackDownload = buildArgs.trackDownload ?: false;
+		var trashed       = IsBoolean( buildArgs.trashed ?: "" ) && buildArgs.trashed;
+		var link          = "";
+		var assetType     = _getAssetManagerService().getAsset(
+			  id           = assetId
+			, selectFields = [ "asset_type" ]
+		).asset_type ?: "";
+
+		if ( !isEmpty( assetType ) ) {
+			trackDownload = !structKeyExists( _getAssetImageType(), assetType );
```
DominicWatson: Rather than this, we could/should perhaps add explicit settings for asset types in `settings.assetmanager.types.{group}.{type}`. Currently we have, for example:

```cfc
types.document.pdf  = { serveAsAttachment=true, mimeType="application/pdf"  };
```

We could allow:

```cfc
types.document.pdf  = { serveAsAttachment=true, mimeType="application/pdf", trackDownloads=true  };
```

And we could default this to the same value as `serveAsAttachment`. Then in the logic here:

```cfc
assetType = assetManagerService.getAssetType( name=assetType );
trackDownload = IsTrue( assetType.trackDownloads ?: ( assetType.serveAsAttachment ) );
```

DominicWatson (changes requested):

system/handlers/core/AssetDownload.cfc
```diff
@@ -21,6 +21,11 @@ component {
 		var config            = assetManagerService.getDerivativeConfig( assetId );
 		var configHash        = assetManagerService.getDerivativeConfigHash( config );
 		var queueEnabled      = isFeatureEnabled( "assetQueue" );
+		var assetPublicUrl    = assetManagerService.getAssetUrl(
```
DominicWatson: We should only do this if it is not a derivative. Should be moved down to the conditional logic around whether this is a derivative, version or plain asset.

I think we are also missing logic for getting version URL?

DominicWatson (commented):
This looks much better :). There's just one little thing I've commented on in the assetDownload handler.

DominicWatson:
Nice one, merged into upcoming 10.14.0 branch.

PR #922, opened 2021-01-13
PRESIDECMS-2048: Catch missing task ex when rendering audit log

smoufa:


DominicWatson:
Nice one thanks @smoufa - releasing in hotfix 10.13.4

PR #923, opened 2021-01-14
PRESIDECMS-2047 exclude deleted form submissions from rule expression

Pixl8RusselCole:
https://presidecms.attlassian.net/browse/PRESIDECMS-2047

Deleted form submission shouldn't include in user submitted forms rule expression


system/services/formbuilder/FormBuilderFilterService.cfc
```diff
@@ -570,6 +570,112 @@ component {
...
+		  required string userId
+		, required array extraFilters
+	) {
+		var websiteUserDao = $getPresideObject( "website_user" );
+
+		var filter = {
+			id = arguments.userId
+	  	}
+
+		result.records = websiteUserDao.selectData(
```
DominicWatson: A cleaner and more efficient approach here is to use `dataExists()` rather than `selectData()`. This will return a `boolean` and select less data from the DB. i.e. this entire function call can be just:

```cfc
return $getPresideObject( "website_user" ).dataExists(
    filter       = { id=arguments.userId }
  , extraFilters = arguments.extraFilters
);
```

DominicWatson (changes requested):
Looking good Russ, nice one. Just one small bit of feedback re running the condition. See comments above.

DominicWatson:
Nice one, merging into upcoming release for QA before releasing in hotfix.

PR #924, opened 2021-01-15
PRESIDECMS-2049 allow partial restriction to apply access condition c…

john5on:
allow partial restriction to apply access condition check


DominicWatson:
Brill, thanks @john5on. Releasing in hotfix 10.13.4

PR #928, opened 2021-01-24
PRESIDECMS-2055 Allow alternative image processor to retrieve image info

john5on:


DominicWatson:
This is good stuff - due to a large refactoring of asset manager file processing, this is not possible to merge as is. That said, this will only now have any effect for custom code manually adding image assets into the asset manager.

The core asset manager upload process, etc. now performs much more efficient code to calculate image sizes and process the images.

PR #930, opened 2021-01-25
PRESIDECMS-2057 secure storage for form builder uploaded files

pixl8-brayden:


system/forms/formbuilder/item-types/fileUpload.xml
```diff
@@ -4,6 +4,7 @@
 		<fieldset id="default">
 			<field name="accept"           control="filetypepicker"  required="false"  sortorder="34"  multiple="true" defaultvalue="" />
 			<field name="maximumFileSize"  control="textInput"       required="false"  sortorder="36"  defaultvalue="5" />
+			<field name="is_private"       control="yesNoSwitch"     required="false"  sortorder="38" />
```
DominicWatson: This isn't quite right - they should *always* be private - should not be a user setting per field.

system/handlers/formbuilder/item-types/FileUpload.cfc
```diff
@@ -9,6 +9,7 @@ component {
 			var downloadLink = event.buildLink(
 				  fileStorageProvider = 'formBuilderStorageProvider'
 				, fileStoragePath     = fileName
+				, fileStoragePrivate  = args.itemConfiguration.is_private ?: false
```
DominicWatson: Should always be `true`

system/handlers/formbuilder/item-types/FileUpload.cfc
```diff
@@ -87,15 +88,25 @@ component {
 	}
 
 	private any function renderResponseToPersist( event, rc, prc, args={} ) {
-		var response = args.response ?: "";
+		var response  = args.response ?: "";
+		var isPrivate = booleanFormat( args.configuration.is_private ?: false );
```
DominicWatson: Should always be `true`

system/i18n/formbuilder/item-types/fileUpload.properties
```diff
@@ -7,4 +7,7 @@ field.maximumfilesize.help=Limit the size of file that your visitors can upload.
 field.accept.title=File types
 field.accept.help=Select the file types that you wish to limit your visitors to uploading. E.g. 'pdf', 'images', etc.
 
+field.is_private.title=Store in private?
```
DominicWatson: Can remove - should not be a user option

DominicWatson (changes requested):
This looks good, thanks @pixl8-brayden . The main thing here however, is that this storage provider should *always* be storing privately - it should not be in user control.

The difficulty with this is how to deal with old files that are in public storage. For this, we will need to add some kind of indicator on form builder file uploads to show that they have been stored privately (so when building the link, we will still be able to work with the old files).

DominicWatson:
Nice one, thanks @pixl8-brayden - merging into 10.15.0 for further testing.

PR #936, opened 2021-02-17
PRESIDECMS-2071 quick add modal add another overlay with form field

pixl8-brayden:


DominicWatson:
Nice one, merged into upcoming 10.15.0 for internal testing before releasing in hotfix.

PR #940, opened 2021-02-22
used common filename if timestamp must be used

philippPSH:
I had problems with saved export files. 
The resulting file path was "uploads/scheduled-export/companies_vps 22-Feb-2021 10:14:13.xlsx". It does not work because of the colon in the file name.

Problem detected on preside 10.12. but it still the source code status in 10.14.


DominicWatson:
Hi @philippPSH thanks for this. Coincidentally, we also were working on a related issue that meant we have fixed this here:

https://presidecms.atlassian.net/browse/PRESIDECMS-2082

The fix we made was to slugify the filename to ensure compatibility with storage providers: https://github.com/pixl8/Preside-CMS/commit/e971f1f3fa72774b9c7b3a360734c8930a1564e9

This means we won't need this PR, but appreciate the work!

PR #941, opened 2021-02-23
PRESIDECMS-2081 formbuilder submissions pagination fix

pixl8-brayden:
Fix formbuilder submissions only show single page result.


DominicWatson:
Nice one, thanks @pixl8-brayden - merged into upcoming release for internal QA before releasing in next hotfix.

PR #942, opened 2021-02-23
PRESIDECMS-2066 event getPageProperty draft & live content fix

pixl8-brayden:


DominicWatson:
This *seems* to me to be in reverse of what is desired. This is using the version tables when only showing live content (no admin logged in or admin choosing to only see live content).

Is that what the intention is here?

DominicWatson:
Nice one, thanks @pixl8-brayden. Merged into the upcoming release branch for internal testing.

PR #944, opened 2021-02-24
PRESIDECMS-2075 MSSQL cascading issue

philippPSH:
Solves PRESIDECMS-2075 even it is not the table in the mentioned error message.

I had the same issue as I started the evaluation and had to fix further three preside object properties. Within my research I just added the code changes in this PR and the other problems disappear.

Tested on MSSQL2016, MariaDB


DominicWatson:
Nice one, thanks @philippPSH - merged into upcoming 10.15.0 release for internal testing (you can find the latest snapshot in Forgebox).

PR #946, opened 2021-02-26
PRESIDECMS-2093 Added feature for datamanager grid to get paged results and total count in single query

jjannek:
* feature disabled by default
* if enabled avoiding second query to obtain total record count in datamanager grid listings
* using sql window function (available in recent mysql/mariadb versions as well as SQL Server. tested with MariaDB)
* especially useful if dealing with large tables but should be in general faster than using 2 queries

### DIFF
--- a/system/services/admin/DataManagerService.cfc
+++ b/system/services/admin/DataManagerService.cfc
@@ -442,18 +442,18 @@ component {
 				args.filterParams[ param.name ] = args.filterParams[ param.name ] ?: param;
 			}
 		}
-		
-		var enhancedGridRecordcount = $isFeatureEnabled( "enhancedGridRecordcount" );
-		
-		if ( enhancedGridRecordcount ) {
-			args.selectFields.append( "count(*) over() as _total_recordcount" );
+
+		var dbAdapter = _getPresideObjectService().getDbAdapterForObject( arguments.objectName );
+
+		if ( dbAdapter.supportsCountOverWindowFunction() ) {
+			args.selectFields.append( "#dbAdapter.getCountOverWindowFunctionSql()# as _total_recordcount" );
 		}
 
 		result.records = _getPresideObjectService().selectData( argumentCollection=args );
 
 		if ( arguments.startRow == 1 && result.records.recordCount < arguments.maxRows ) {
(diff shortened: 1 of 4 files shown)


system/services/admin/DataManagerService.cfc
```diff
@@ -442,11 +442,19 @@ component {
 				args.filterParams[ param.name ] = args.filterParams[ param.name ] ?: param;
 			}
 		}
+		
+		var enhancedGridRecordcount = $isFeatureEnabled( "enhancedGridRecordcount" );
+		
+		if ( enhancedGridRecordcount ) {
+			args.selectFields.append( "count(*) over() as _total_recordcount" );
```
DominicWatson: is this ANSI SQL? compat with SQL server, MySQL and Postgres?

jjannek: I tested it in Mariadb and SQL Server. according to the interweb Postgres should work as well. It depends on the dbms engine version when the vendor introduced the functionality.
They are called window functions and Mariadb for instance supports count in window functions since version 10.2.

DominicWatson: Right kk, cool. I think if we're going to spend time at all on this, let's get it tested and working in the upcoming release branch. I'd suggest:

1. DB adapter method that signals support for the feature (defaults to false and can be dynamic - but the result cache for once in the lifetime of the app)
2. Ditch the feature flag
3. Add DB adapter method for getting the SQL to use (`"#sqlToUse# as _total_recordcount"`)

sebduggan: I've just tested and can confirm that the syntax works in PostreSQL too...

sebduggan: Unsupported engines (if there are any) could simply return NULL as the sqlToUse, and that would trigger a separate count query via the current method.

jjannek: Changes implemented as discussed. I decided to go with Dom's original idea and added an additional check method on the adapter in order to not have the extra select field added even if db does not support it.

DominicWatson: Nice - merging into upcoming release now.

PR #948, opened 2021-03-01
PRESIDECMS-2022 search and filter using formula fields

karhong:


DominicWatson:
Hey @karhong - this looks really great. However, I think it would make sense to split this into two separate pieces of work (sorry to be a pain).

1. Searchable enums directly in the data table
2. Formula field expressions for filters

Would you be able to back out one of the changes from this PR, and make a new ticket + PR with just those changes?

karhong:
Hi @DominicWatson, I had created another PR specifically for enum fields.
https://github.com/pixl8/Preside-CMS/pull/949

system/services/rulesEngine/RulesEngineAutoPresideObjectExpressionGenerator.cfc
```diff
@@ -49,17 +49,14 @@ component {
...
-		}
 
 		var isRequired   = IsBoolean( propertyDefinition.required ?: "" ) && propertyDefinition.required;
 		var propType     = propertyDefinition.type ?: "string";
 		var relationship = propertyDefinition.relationship ?: "";
 		var relatedTo    = propertyDefinition.relatedTo ?: "";
 		var expressions  = [];
 
-		if ( !isRequired && !( [ "many-to-many", "one-to-many" ] ).findNoCase( relationship ) ) {
+		if ( !isRequired && !( [ "many-to-many", "one-to-many" ] ).findNoCase( relationship ) && !Len( Trim( propertyDefinition.formula ?: "" ) ) ) {
```
sebduggan: Can we set an `isFormula` variable once at the top of the method and then use that throughout?

DominicWatson:
Great work. Merged into upcoming release.

system/services/presideObjects/PresideObjectService.cfc
```diff
@@ -2189,6 +2190,53 @@ component displayName="Preside Object Service" {
 		return fields;
 	}
 
+	public array function parseSelectFieldsForHavingClause(
```
DominicWatson: I think the name of this function is a little confusing/misleading. Could this be changed to `expandHavingClauses()`?

system/services/presideObjects/PresideObjectService.cfc
```diff
@@ -2189,6 +2190,53 @@ component displayName="Preside Object Service" {
...
+
+					if ( propertyLen == 1 ) {
+						if( !ArrayFindNoCase( selectFields, field ) ) {
+							field = "#arguments.objectName#.#field#";
+						}
+					}
+					else if (propertyLen == 2 ) {
+						var prefix       = ListFirst( field, "." );
+						var propertyName = ListLast( field, "." );
+						var prop         = props[ propertyName ] ?: {};
```
DominicWatson: I believe there may be a strong possibilty that the property does not belong to the object directly. i.e. it could be `related$object$something.propertyName`.

We have similar logic for getting the property from source object and field path in this service and we should re-use it here to ensure we are reading the property from the correct place.

DominicWatson (changes requested):
Looking good, still some work to do here though I think

DominicWatson:
Nice one, merged into upcoming 10.15 for further testing

PR #949, opened 2021-03-02
PRESIDECMS-2085 search and filter using enums fields

karhong:


system/services/admin/DataManagerService.cfc
```diff
@@ -787,7 +797,11 @@ component {
 				}
 
 				filter   &= " )";
-				termDelim = " and ";
+				if ( arrayLen( objEnumFields ) ) {
+					termDelim = " or ";
+				} else {
+					termDelim = " and ";
+				}
```
sebduggan: This doesn't really work: it completely changes the way multiple word searches work.

Currently, if your search term is "seb duggan", the "and" term delimiter means that it matches records with BOTH "seb" AND "duggan". But this logic changes it so, if there are enums in the object, it will match EITHER "seb" OR "duggan". So, the termDelim should ALWAYS be "and".

The logic should move up inside the parsedFields loop (after line 795), so if the current field in the loop is an enum, it looks to see if the search term is an enum label _for that enum_, and if it is then we add in something like:

`filter &= " or " & field & " = :#enumParamName#";`

(we'd also need to create a separate index of enum params which would get added to params later in the method).

Note also that the term might match more than one enum label, so we'd have to handle that too.

system/services/admin/DataManagerService.cfc
```diff
@@ -766,6 +769,13 @@ component {
 		var poService            = _getPresideObjectService();
 		var relationshipGuidance = _getRelationshipGuidance();
 		var searchTerms          = arguments.expandTerms ? listToArray( arguments.q, " " ) : [ arguments.q ];
+		var objEnumFields        = poService.getObjectEnumProperties( objectName=arguments.objectName );
```
sebduggan: This checks if an object has enum properties and gets them all - whether or not they are included in the fields to be searched.

See comment below for an alternative way to check the enum fields.

system/services/admin/DataManagerService.cfc
```diff
@@ -813,8 +816,25 @@ component {
...
 						if ( _propertyIsSearchable( field, objName ) ) {
-							filter &= delim & fullFieldName & " like :#paramName#";
-							delim = " or ";
+							var fieldEnumName = poService.getObjectPropertyAttribute(
+								  objectName    = objName
+								, propertyName  = field
+								, attributeName = "enum"
+							);
+
+							if ( !isEmpty( fieldEnumName ) ) {
```
sebduggan, line 845: This enum match should be done _in addition to_ the normal LIKE comparison if it's an enum, not instead of.

So the resulting SQL would be:

... fullFieldname LIKE :param OR fullFieldName = :enumMatch ...

And ideally we'd parameterise the enumMatch value.

DominicWatson:
Nice one, merging this for QA.

PR #950, opened 2021-03-02
Added a fix to data-manager when using a formula label field.

cfmitrah:
https://presidecms.atlassian.net/browse/PRESIDECMS-1674

### DIFF
--- a/system/services/presideObjects/PresideObjectService.cfc
+++ b/system/services/presideObjects/PresideObjectService.cfc
@@ -2190,7 +2190,7 @@ component displayName="Preside Object Service" {
 					objName = _resolveObjectNameFromColumnJoinSyntax( startObject=arguments.objectName, joinSyntax=ListFirst( match, "." ) );
 				}
 
-				labelField = getObjectAttribute( objName, "labelfield", "label" );
+				labelField = listLast(getObjectAttribute( objName, "labelfield", "label" ));
 				if ( !Len( labelField ) ) {
 					throw( type="PresideObjectService.no.label.field", message="The object [#objName#] has no label field" );
 				}


DominicWatson:
Thanks for this @cfmitrah . Before I merge however, could you provide an example value for this that this fix would address? I'm not sure why `ListLast()` is going to solve it?

cfmitrah:
Hi @DominicWatson , By Default when a label has more than one value in data-manager, Only last field will be considered as label. A single formula-field as label do not cause error. But Including a formula-field in the multiple labels causes error in select-fields. To ensure including formula-field do not cause error, Listlast function helps to get last field in the list of labels.

DominicWatson:
I'm still not sure what this is. `@labelField` should never be a list of multiple fields...? Can you give an example of what value this is accounting for?

cfmitrah:
Hi @DominicWatson 
```
/**
 * @labelfield 	field1, field2, field3
 **/
component {
	property name="field1" type="string" ;
	property name="field2" type="string" ;
	property name="field3"  formula="concat( field1 , ' ' , field2)";
}
```


If the label annotation has 2 or more values(ie field1, field2, field3 ) with field3 as formula-field, then it gives the error.

*Error: Unknown column 'field3' in 'field list'*

By Default when a label has more than one value in data-manager, Only last field will be considered. 
A single formula-field as label do not cause error. To ensure including formula-field do not cause error, Listlast function helps to get last field in the list of labels.

DominicWatson:
I see. I'm pretty sure that having a _list_ of label fields is not supported in Preside in any way. Perhaps it works in some scenarios but there isn't any designed features to work with this.

What would the purpose be of setting a list of fields for labelfield?

cfmitrah:
There is no purpose for having list of multiple fields in labelfield. The single labelfield with formula is already handled, But giving multiple label fields with formula based on preside object throw an error on grid action.

DominicWatson:
I think I'm going to leave this one, but thank you @cfmitrah. Has potential to break something without a clear benefit. If you've put a list of fields in as the label field, then expect it to break in unexpected ways.

PR #951, opened 2021-03-02
PRESIDECMS-2092 Fix incorrect expression when concatenating two strings

smoufa:


DominicWatson:
Great, thanks for this @smoufa . Merging into upcoming release but will make its way into the next hotfix.

PR #954, opened 2021-03-03
PRESIDECMS-1574 object picker prefilled with deleted value

sebduggan:


system/handlers/formcontrols/ObjectPicker.cfc
```diff
@@ -64,4 +66,23 @@ component {
...
 		return renderView( view="formcontrols/objectPicker/index", args=args );
 	}
+
+	private string function _removeInvalidValues( required string objectName, required string values ) {
+		if ( !len( args.defaultValue ?: "" ) ) {
+			return "";
+		}
+
+		var initialValues = listToArray( arguments.values );
+		var validValues   = presideObjectService.selectData(
```
DominicWatson, line 76: I _think_ this can just be:

```cfc
var existingValues = ...selectData(...);

return ColumnData( existingValues.id );
```

???

sebduggan: Potentially multiple, ordered values. This keeps them in the original order...

DominicWatson: ah ok, yes.

PR #955, opened 2021-03-03
PRESIDECMS-2088 update object picker cachebuster when records deleted

sebduggan:


system/services/admin/DataManagerService.cfc
```diff
@@ -664,6 +664,7 @@ component {
 		var dmField           = obj.getDateModifiedField();
 		var lastModified      = Now();
 		var rendererCacheDate = _getLabelRendererService().getRendererCacheDate( labelRenderer );
+		var recordCount       = obj.selectData( recordCountOnly=true );
```
DominicWatson: Any way to improve efficiency by having both count + max(datemodified) in single query? Defo only need to get recordcount when we have date modified field though (because we always bust the cache when no dm field).

PR #956, opened 2021-03-04
Added a fix to object picker with searching some characters

cfmitrah:
https://presidecms.atlassian.net/browse/PRESIDECMS-1405

### DIFF
--- a/system/assets/js/admin/presidecore/preside.uber.select.js
+++ b/system/assets/js/admin/presidecore/preside.uber.select.js
@@ -267,10 +267,10 @@
 
 		UberSelect.prototype.winnow_results = function() {
 			var $uberSelect = this
-			  , searchText  = $uberSelect.get_search_text();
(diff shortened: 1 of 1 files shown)


DominicWatson:
Nice, thanks @cfmitrah . I'm wondering a couple of things:

1. Could we override `get_search_text` on the uberselect object so that we only have to do this logic once?
2. Is there a more generic xml decoding function we could use to decode the search string? This fixes three chars but logically it sounds like there could be more?

cfmitrah:
Hi @DominicWatson , 
	All the other characters works fine except the three(&,<,>) and I could use the logic once. But, I found a better solution for the issue. Here's the fix. In this, the input will be received as html and compared as text. I am sure, this fix will work for all scenarios.

DominicWatson:
Aha, very tidy! Will get this merged into upcoming 10.15.0 release and tested. Thank you.

PR #957, opened 2021-03-04
Added a fix to overlay of text content in Data manager

cfmitrah:
https://presidecms.atlassian.net/browse/PRESIDECMS-2089


DominicWatson:
Nice one, thanks @cfmitrah . Merged into upcoming release for further testing.

PR #958, opened 2021-03-10
PRESIDECMS-2094 introduce permission handlers config setting

pixl8-brayden:


system/services/websiteUsers/WebsitePermissionService.cfc
```diff
@@ -87,6 +90,24 @@ component displayName="Website permissions service" {
...
 	) {
+		var coldbox            = $getColdbox();
+		var permissionHandlers = _getPermissionHandlers();
+
+		for ( var permHandler in permissionHandlers ) {
+			if ( !isEmpty( permHandler.handler ?: "" ) && coldbox.viewletExists( permHandler.handler ) ) {
+
+				for ( var key in ( permHandler.keyPatterns ?: [] ) ) {
+					if ( reFindNoCase( key, arguments.permissionKey ) ) {
+						return coldbox.renderViewlet(
```
DominicWatson: This should not be a _viewlet_. A viewlet is something for rendering content. In this case, we would always expect this to be a handler, use:

```cfc
if ( $getColdbox().handlerExists( permHandler.handler ) ) {
    var result = $runEvent( 
          event          = permHandler.handler
        , eventArguments = { args=arguments } 
        , private        = true
        , prePostExempt  = true
    );
    
    // this, or raise a helpful error if the returned value is not a boolean
    return $helpers.isTrue( local.result ?: "" ); // it may be null
}

File: system/config/Config.cfc [hunk 2]
```diff
@@ -366,7 +366,8 @@ component {
 		settings.adminRoles.savedExportManager = [ "cms.access", "savedExport.*" ];
 		settings.adminRoles.savedExportAccess  = [ "cms.access", "savedExport.navigate", "savedExport.read" ];
 
-		settings.websitePermissions = {
+		settings.websitePermissionHandlers = [];
```
DominicWatson: It might be worth exploring having a handler _convention_ here rather than needing to configure them in Config.cfc. i.e. `/handlers/websitepermissions/SomeHandler.cfc$checkPermission()`. The keys could then be annotated on the method (like how we do rules engine expressions):

```cfc
// handlers/websitepermissions/CustomPermissionHandler.cfc
component {

    /**
     * @keyPatterns groups.*
     */
    private boolean function checkPermission( event, rc, prc, args={} ) {
        // custom logic here
        return false;
    }
}
```

DominicWatson (commented):
Looking good. I've put some feedback in review for your consideration.

system/services/websiteUsers/WebsitePermissionService.cfc
```diff
@@ -94,17 +96,17 @@ component displayName="Website permissions service" {
...
-				for ( var key in ( permHandler.keyPatterns ?: [] ) ) {
-					if ( reFindNoCase( key, arguments.permissionKey ) ) {
-						return coldbox.renderViewlet(
-							  event = permHandler.handler
-							, args  = arguments
-						);
-					}
-				}
-
+			if ( $getColdbox().viewletExists( permissionHandlers[ permHandler ].handler ) &&
```
DominicWatson: This still needs to be `handlerExists()` and not `viewletExists()`.

DominicWatson (changes requested):
Looks great, just one tiny thing to fix up.

DominicWatson:
Nice one, thanks @pixl8-brayden - this is merged into upcoming 10.15.0 release for further QA, etc.

PR #959, opened 2021-03-12
Added a fix to the issue "Email Links in emails lead to JS being generated by default"

cfmitrah:
https://presidecms.atlassian.net/browse/PRESIDECMS-2015


DominicWatson:
Nice one, thanks @cfmitrah - have merged into upcoming release for internal QA.

PR #960, opened 2021-03-19
Optimise directory listing approach for resourceBundles

zspitzer:


DominicWatson:
Thanks for this @zspitzer + ensuing discussion. As per discussion, I've looked at an alternative approach that does not do directory list at all (because we have already done this directorylist lookup on startup). This appears to make a pretty big difference (especially on slower file systems):

https://github.com/pixl8/Preside-CMS/commit/c1d3c382046ffa4d4ddca631cc5a605e9d24f36a

Will be available in a 10.15.0-SNAPSHOT to play with once this build completes: https://github.com/pixl8/Preside-CMS/runs/2150464016?check_suite_focus=true

PR #963, opened 2021-03-24
PRESIDECMS-2116 avoid unnecessary exceptions

zspitzer:
Both `_getUseCacheDefault()` and `_getDefaultAllowDraftVersions` in `PresideObjectService` use try/catch to see whether or not this has already been calculated for the request. A simple `StructKeyExists()` should suffice here and perform better.

https://presidecms.atlassian.net/browse/PRESIDECMS-2116

### DIFF
--- a/system/services/presideObjects/PresideObjectService.cfc
+++ b/system/services/presideObjects/PresideObjectService.cfc
@@ -3649,21 +3649,15 @@ component displayName="Preside Object Service" {
 	}
 
 	private boolean function _getUseCacheDefault( required string objectName ) {
-		try {
-			return request[ "_defaultUseCache#arguments.objectName#" ];
-		} catch( any e ) {
+		if ( !StructKeyExists( request, "_defaultUseCache#arguments.objectName#" ) )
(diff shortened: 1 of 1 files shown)


DominicWatson:
Nice one, thanks Zac. Not sure why these are using try/catch here - would certainly tend to follow the structkeyexists pattern in general.

One little thing, and I can fix this when merging this, our internal styleguide is to not use single line IF statements. So:

```cfc
if ( something ) {
    // single line
}

// NOT
if ( something ) 
    // single line
```

(I'll fix that). I'll also create a ticket for this. Presumably the purpose of this is that more processing/memory consumption used when try/catch is being used?

zspitzer:
More that it was pointless than anything else, sure it's a bit more efficient, but there was simply no need for it!

I'll read the style guide

DominicWatson:
Nice one - nothing to read publicly. Its usually just little things that we ignore and tweak when merging, all good. But thatnks for changing anyway. Will get this merged in next batch of PR merging some time v soon.

PR #965, opened 2021-03-25
Fixed the drop-down overlap issue on quick add modal window

cfmitrah:
https://presidecms.atlassian.net/browse/PRESIDECMS-2115


DominicWatson:
Great, thanks @cfmitrah - looking good. Will get tested in upcoming release.

PR #967, opened 2021-03-29
PRESIDECMS-2119 preside email css inliner tool injects media query styles

nickcervantes:
I used the regex replace all to remove all media query clauses.

Regexp: `\@media.*?\{(.*?\})?\s*?\}`
- Finding strings starting with `@media` up to the first occurrence of '{' and until the occurrence of consecutive '}'. Consecutive '}' means it is the end of the media query clause.
- The first '}' was put inside a group, `(.*?\})?`, to capture the scenario of empty media query, eg. `@media { }`

### DIFF
--- a/system/services/email/EmailStyleInliner.cfc
+++ b/system/services/email/EmailStyleInliner.cfc
@@ -51,7 +51,7 @@ component {
 		var styles        = [];
 
 		for( var styleElement in styleElements ) {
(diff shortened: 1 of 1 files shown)


DominicWatson:
Fab, thanks @nickcervantes - merging into upcoming 10.15.0 release.

PR #971, opened 2021-04-06
PRESIDECMS-2161 variable scoping

zspitzer:


DominicWatson:
Nice one, thanks @zspitzer - merged into upcoming release

PR #973, opened 2021-04-19
PRESIDECMS-2126 Added separate resize crop transformation for derivative

jtcy0123:
Added a separate resize crop transformation for derivative to render only the crop areas on FE

### DIFF
--- a/system/services/assetManager/ImageManipulationService.cfc
+++ b/system/services/assetManager/ImageManipulationService.cfc
@@ -38,15 +38,17 @@ component displayname="Image Manipulation Service" {
 		,          string  cropHint            = ""
 		,          string  useCropHint         = false
 		,          struct  fileProperties      = {}
+		,          string  ignoreDimension     = false
 	) {
 		var args = arguments;
 
 		if ( arguments.useCropHint && arguments.cropHint.len() ) {
 			args.cropHintArea = _getCropHintArea(
-				  image    = args.asset
-				, width    = args.width
-				, height   = args.height
-				, cropHint = args.cropHint
+				  image           = args.asset
+				, width           = args.width
+				, height          = args.height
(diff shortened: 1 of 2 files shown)


system/services/assetManager/ImageManipulationService.cfc
```diff
@@ -38,15 +38,17 @@ component displayname="Image Manipulation Service" {
 		,          string  cropHint            = ""
 		,          string  useCropHint         = false
 		,          struct  fileProperties      = {}
+		,          string  ignoreDimension     = false
```
DominicWatson: Is there any reason this should not be a `boolean` type?

system/services/assetManager/ImageManipulationService.cfc
```diff
@@ -38,15 +38,17 @@ component displayname="Image Manipulation Service" {
...
 			args.cropHintArea = _getCropHintArea(
-				  image    = args.asset
-				, width    = args.width
-				, height   = args.height
-				, cropHint = args.cropHint
+				  image           = args.asset
+				, width           = args.width
+				, height          = args.height
+				, cropHint        = args.cropHint
+				, ignoreDimension = $helpers.isTrue( args.ignoreDimension )
```
DominicWatson: As above, if the argument is a `boolean` type, then no need for the `isTrue()` helper.

DominicWatson:
Looks great @jtcy0123 thanks for this. I've a tiny bit of feedback on the logic. Could you take a look and also provide a clear example in the JIRA ticket for future reference (helps to describe what's in our next release).

DominicWatson (changes requested):

DominicWatson:
Nice one, merging into upcoming 10.15.0 release for further QA

PR #974, opened 2021-04-21
PRESIDECMS-2128 datamanager add one to many action customization

karhong:


system/handlers/admin/DataManager.cfc
```diff
@@ -2423,6 +2423,16 @@ component extends="preside.system.base.AdminHandler" {
 
 		validationResult = validateForm( formName=arguments.formName, formData=formData, stripPermissionedFields=arguments.stripPermissionedFields, permissionContext=arguments.permissionContext, permissionContextKeys=arguments.permissionContextKeys );
 
+		args.formData         = formData;
```
DominicWatson, line 2427: `args` has not been initialized here. ie, we should probably have it in the arguments list ( `struct args = {}`)

DominicWatson (changes requested):

DominicWatson:
Looking good, just one little thing to feed back on.

system/handlers/admin/DataManager.cfc
karhong: I'm initialized via `var args = arguments;` by following how the DataManager._addRecordAction did instead. Might also useful to have more params passing to customization point

DominicWatson:
Nice one, merged into upcoming release.

PR #976, opened 2021-04-21
PRESIDECMS-2130 Formbuilder error when missing item type

teonator:

PR #979, opened 2021-04-23
PRESIDECMS-2130 Formbuilder item type not found

teonator:


DominicWatson:
Removing as is a duplicate of https://github.com/pixl8/Preside-CMS/pull/976

PR #980, opened 2021-04-26
Feature presidecms 2135 allow js confirmation open new window

john5onCheng:


DominicWatson:
Great stuff, merging into upcoming release for QA.

DominicWatson:
@john5onCheng - just doing some testing on this and can see that links _do_ open in a new window/tab. However, when you return to the original tab, the dialog is still open. In the case where the link target is not the default same window behaviour, we should close the dialog at this point also.

john5onCheng:
Hi @DominicWatson ,

I have added the code to close dialog.

DominicWatson:
Nice one thanks @john5onCheng  - merging ready for retest

PR #981, opened 2021-04-26
PRESIDECMS-2137 Disk store cache reset

teonator:


DominicWatson:
Nice one thanks @teonator . Merging into upcoming release for testing but will get into hotfix asap

PR #983, opened 2021-04-27
PRESIDECMS-2136 Show non live content

teonator:


system/coldboxModifications/RequestContextDecorator.cfc
```diff
@@ -358,7 +358,11 @@ component accessors=true extends="preside.system.coldboxModifications.RequestCon
 
 	public boolean function showNonLiveContent() {
 		try {
-			return request._showNonLiveContent;
+			if ( this.isAdminUser() ) {
+				return true;
```
DominicWatson: This negates the purpose of the feature. As an admin, you can choose whether or not to show non-live content in the Preside admin bar (when viewing the frontend).

The issue here is that when the _default_ is set to _not_ show non-live content in `Config.cfc`, it starts also effecting the backend (i.e. data manager)

DominicWatson (changes requested):

DominicWatson:
Aha, nice find :) Merging into 10.15 for testing.

PR #985, opened 2021-04-30
Added a fix to the issue 'Faulty quick edit in admin object select'

cfmitrah:
https://presidecms.atlassian.net/browse/PRESIDECMS-1351


DominicWatson:
Interesting fix! Thanks @cfmitrah - will get this in QA in our upcoming release.

PR #991, opened 2021-05-21
Feature presidecms 2151 add csrf token to form builder form

nodoherty:


system/handlers/formbuilder/Core.cfc
```diff
@@ -13,6 +13,11 @@ component {
 			event.notFound();
 		}
 
+		if ( !event.validateCsrfToken( rc.csrfToken ?: "" ) ) {
+			var csrfErrorMessage = translateResource( uri="formbuilder.item-types.csrf_token:invalid.csrf_token.error.message" );
+			setNextEvent( url=cgi.http_referer, persistStruct={ errorMessage=csrfErrorMessage } );
```
DominicWatson: We should probably be persisting the submission here also so that the user does not lose any data that they have filled in. i.e.

```cfc
var submission = event.getCollectionWithoutSystemVars(); // (raised from lines below)
// ...
var persist = submission;
persist.errorMessage = csrfErrorMessage;
// ...
```
```

File: system/i18n/formbuilder/item-types/csrf_token.properties [hunk 2]
```diff
@@ -0,0 +1,5 @@
+title=CSRF Token
```
DominicWatson: This isn't an item type really - so suggest putting that error message somewhere general like `cms.properties`.

system/i18n/formbuilder/item-types/csrf_token.properties
```diff
@@ -0,0 +1,5 @@
...
+iconclass=fa-shield red
+description=CSRF token protection
+
+invalid.csrf_token.error.message=Invalid CSRF Token. Please try submitting your form again.
```
DominicWatson: This message should be less techy - I think we should have examples of CSRF messages elsewhere that we can use. But basically something that doesn't mention "CSRF". Perhaps:

> Your form submission has been prevented due to a security check failure. This may be caused by a timeout. Please review your form entries and resubmit.

^^ something _like_ that.

DominicWatson (changes requested):
Looking good, thanks @nodoherty. Have some feedback to review.

system/handlers/formbuilder/Core.cfc
nodoherty: Thanks @DominicWatson  - the `getCollectionWithoutSystemVars()` method removes the `csrfToken` value before returning but I noticed it does a `duplicate()` beforehand so should be safe!!

Updated.

system/i18n/formbuilder/item-types/csrf_token.properties
nodoherty: Found cms:invalidCsrfToken.error in the core so updating to use that instead.

system/i18n/formbuilder/item-types/csrf_token.properties
nodoherty: Updated to what was already in core!

DominicWatson:
:boom: looking good. Will get merged into upcoming for some QA.

PR #995, opened 2021-06-03
PRESIDECMS-2155 allow objectpicker to specify target id field

s-ven-315:


system/services/admin/DataManagerService.cfc
```diff
@@ -612,7 +612,7 @@ component {
 		if (args.orderBy is 'label') {
 			args.orderBy = labelField;
 		}
-		var idField            = _getPresideOBjectService().getIdField( arguments.objectName );
+		var idField            = arguments.idField ?: _getPresideOBjectService().getIdField( arguments.objectName );
```
DominicWatson: * `idField` should be declared in the list of arguments to this function (and can default to an empty string)
* in `system/handlers/admin/DataManager.cfc#900`, the value of this argument is passed in as `""` in some scenarios. We need to defend against an empty IDField here. So:

```
var idField = Len( Trim( arguments.idField ) ) ? arguments.idField : _getPresideOBjectService().getIdField( arguments.objectName );
```

DominicWatson (changes requested):
Thanks for this @s-ven-315 , looking good. Just a little bit of feedback on the service method. Could you take a look?

system/services/admin/DataManagerService.cfc
s-ven-315: Added idField to argument list and updated code as suggested. Thanks!

DominicWatson:
Great, thanks @s-ven-315 . Merging into upcoming release. For QA, etc.

PR #1000, opened 2021-06-11
PRESIDECMS-2157 custom export url argument for object data table

pixl8-brayden:


DominicWatson:
Nice one, thanks @pixl8-brayden - will get this merged and tested for upcoming 10.15.0 release.

PR #1001, opened 2021-06-11
avoid setTimeout unless necessary

zspitzer:


DominicWatson:
Thanks for this @zspitzer - this may require some further dev and testing. What might be tricky is that the targeted class for datatables does not exist in the DOM until after the jQuery datatables library has done initializing the table(s). 

So we may need to delay the check for some sensible amount of time and then stop doing repeated checks if no tables found.

cfmitrah:
Correct way could be using proper event ( fnInitComplete [ ref : http://legacy.datatables.net/usage/#fnInitComplete ] or initComplete [ ref : https://datatables.net/reference/option/initComplete ] ) while initialize dataTable against a DOM table, Simple code search gave below 3 places where dataTables are initialized

https://github.com/pixl8/Preside-CMS/blob/stable/system/assets/js/admin/specific/notifications/notificationTable.js#L34
https://github.com/pixl8/Preside-CMS/blob/stable/system/assets/js/admin/presidecore/preside.static.datatable.js#L10
https://github.com/pixl8/Preside-CMS/blob/stable/system/assets/js/admin/frontend/frontendEditors.js#L448

PR #1005, opened 2021-07-09
PRESIDECMS-2170 Admin toolbar impersonate user shortcut & logout

teonator:


DominicWatson:
Nice enhancement, thanks @teonator - will merge into upcoming release. FYI, am removing the `target="_blank"` here.

PR #1006, opened 2021-07-12
PRESIDECMS-2172 Rule engine onetomanyhas required

teonator:


DominicWatson:
Brill, nice find @teonator - merging into upcomning release for testing but will hotfix once through QA.

PR #1007, opened 2021-07-13
PRESIDECMS-2174 preview email and sent email potentially rendering differently

karhong:


system/services/email/EmailTemplateService.cfc
```diff
@@ -146,7 +146,7 @@ component {
 				, unsubscribeLink    = unsubscribeLink
 				, viewOnline         = viewOnline
 			);
-			message.htmlBody = $renderContent( renderer="richeditor", data=preppedHtml.html, context="email", args={ styles=preppedHtml.styles } );
```
DominicWatson: It is important to do the `$renderContent` _after_ `_prepareHtml` because `_prepareHtml` can be _cached_ without dynamic content. We then render richeditor content which can add dynamic content.

I think this will need revisiting.

DominicWatson (changes requested):
Thanks for this @karhong - its not quite the right approach, however. If anything, I think it will need to be the previewEmail logic that will need updating.

DominicWatson:
Have refactored differently - instead, we just proxy previewTemplate to prepareParameters() with some different options to ensure we are in preview mode. Thanks for the input.

PR #1014, opened 2021-08-05
PRESIDECMS-2184 Missing question when form cloned

teonator:
- Fix question not cloned.


DominicWatson:
Nice one, thanks @teonator - merging into upcoming release for QA.

PR #1016, opened 2021-08-23
PRESIDECMS-2190 permissions for data manager export fields

pixl8-brayden:
To be use along with object properties' attribute `limitToAdminRoles`, and the value could be a list of admin roles.

### DIFF
--- a/system/handlers/formcontrols/MultiSelectPanel.cfc
+++ b/system/handlers/formcontrols/MultiSelectPanel.cfc
@@ -1,5 +1,7 @@
 component {
 	property name="presideObjectService" inject="PresideObjectService";
+	property name="loginService"         inject="LoginService";
+	property name="permissionService"    inject="PermissionService";
 
 	public string function index( event, rc, prc, args={} ) {
 		args.labels    = args.labels ?: [];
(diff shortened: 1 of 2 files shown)


system/handlers/formcontrols/MultiSelectPanel.cfc
```diff
@@ -14,7 +16,20 @@ component {
 				var props   = presideObjectService.getObjectProperties( objectName );
 
 				for ( var prop in props ) {
-					if ( !( props[ prop ].relationship ?: "" ).reFindNoCase( "to\-many$" ) && !IsTrue( props[ prop ].excludeDataExport ?: "" ) ) {
+					var hasPermission     = true;
```
DominicWatson: We could/should tweak this to improve performance I think. We _only_ need to do the permissions checking _after_ first doing the straightforward check:

```
if ( !( props[ prop ].relationship ?: "" ).reFindNoCase( "to\-many$" ) && !IsTrue( props[ prop ].excludeDataExport ?: "" ) ) {
...
```

i.e. keep the existing check as it was and then run the additional permission checking within that. Otherwise we end up permission checking fields that would never have made it anyway.

system/services/security/PermissionService.cfc
```diff
@@ -225,6 +225,59 @@ component displayName="Admin permissions service" {
...
+		var userGroups  = listUserGroups( argumentCollection=arguments );
+
+		if ( arrayLen( userGroups ) ) {
+			var rolesQuery = _getGroupDao().selectData(
+				  filter       = { id=userGroups }
+				, selectFields = [ "roles" ]
+			);
+
+			for ( var q in rolesQuery ) {
+				for ( var role in q.roles ) {
```
DominicWatson: Should this be `ListToArray( q.roles )`? It is unclear how we're looping over this field from looking at the code.

DominicWatson (changes requested):
Looking good @pixl8-brayden - nice work. A tiny bit of feedback to review before merging and testing.

PR #1017, opened 2021-08-23
PRESIDECMS-2191 save export permission

pixl8-brayden:
ability to control data manager listing export feature with `dataManagerSaveExportPermissionKey` annotation, default to `read`


DominicWatson:
Looking good :) merging into upcoming release for QA.

PR #1018, opened 2021-08-24
PRESIDECMS-2192 send test email is ignoring draft version attachment

mookeshNanda:
Added `allowDrafts` & `fromVersionTable` param for `getAttachments()` function, when then will be used on selectData param based on the arguments.

It is similar approach as current getTemplate() function.


DominicWatson:
This looks great, thanks @mookeshNanda - looks like the ticket got dropped somewhere along the line so apologies for the slow response. Merging for QA.

PR #1020, opened 2021-08-26
PRESIDECMS-2195 Admin homepage

teonator:
Added new admin user homepage.


system/views/admin/layout/userMenu.cfm
```diff
@@ -15,6 +17,12 @@
 	</a>
 
 	<ul class="user-menu dropdown-menu dropdown-yellow dropdown-caret dropdown-close">
+		<li>
+			<a href="#event.buildAdminLink( linkTo="homepage.make", queryString="data=#data#" )#" >
+				<i class="fa fa-home"></i>
+				#translateResource( "cms:homepage.make.link" )#
+			</a>
+		</li>
```
sebduggan: I'd suggest that there should be logic around this so it only shows for admin requests.

So, if you are logged in as an admin user, and are viewing a front-end page (and so seeing the admin meu bar), there should be no option to make the front-end page your admin homepage...

sebduggan (changes requested):

system/views/admin/layout/userMenu.cfm
teonator: Agree. Will hide the option from front-end admin bar.

DominicWatson:
Awesome sauce - merged into upcoming 10.16.0 release.

sebduggan (approved):

system/interceptors/CsrfProtectionInterceptor.cfc
```diff
@@ -22,7 +22,7 @@ component extends="coldbox.system.Interceptor" {
 				setNextEvent( url=cgi.http_referer, persistStruct=persist );
 			}
 
-			setNextEvent( url=event.buildLink( linkTo=applicationsService.getDefaultEvent() ), persistStruct=persist );
+			setNextEvent( url=event.buildLink( linkTo=applicationsService.getDefaultEvent(). queryString=applicationsService.getDefaultQueryString() ), persistStruct=persist );
```
DominicWatson, line 25: I think you have ` .` here instead of an `,` in the arguments list to `event.buildLink()`

system/views/admin/layout/userMenu.cfm
```diff
@@ -4,6 +4,8 @@
 	} else {
 		logoutUrl = event.buildAdminLink( linkTo="login.logout", queryString="redirect=referer" );
 	}
+
+	data = EncodeForUrl( SerializeJson( event.getContext() ) );
```
DominicWatson: I think this is not right. Rather than grabbing the entirety of the `rc` - we could/should just grab `url` explicitly? We really only care about what is in the current URL.

DominicWatson (changes requested):
Just looking through this for the first time and there are a few things here to pick up on. However, overall, I think this could be greatly simplified:

1. Create a new requestContextDecorator method, `getAdminHomepageUrl()` - this should then be used in all the places that are currently doing any homepage link building
2. When setting the homepage, we should just set the URL, rather than a serialized struct of the rc. `event.getCurrentUrl()` is perfect for this
3. The input should be validated when setting the homepage data - we can validate it by checking that it _starts with_ `/#settings.preside_admin_path#/`
4. In the admin route handler, if it detects that the URL is just `/admin/` - it should check the admin homepage URL - if it is not the same, then it should _redirect_ - rather than trying to serve the homepage dynamically on that URL.

PR #1021, opened 2021-08-27
PRESIDECMS-2196 object property attribute to exclude auto generate filter keys

pixl8-brayden:
The cached expression files will have different suffix when the login user is not a system user (new suffix will be hash of exclude tags + logged in admin user's roles).

Each property in an object able to exclude expression keys to be generate by new `excludeAutoExpressions` attribute.
Developer also able to control auto generate expression keys available for specific admin roles by new `autoFilterExpressions:{admin role}` attribute.


DominicWatson:
Nice one @pixl8-brayden - going to merge this ready for QA in upcoming release branch (10.16)

PR #1026, opened 2021-09-09
Feature presidecms 2204 formbuilder v2 allows multiple of same question

Pixl8RusselCole:


DominicWatson:
Nice one, thanks @Pixl8RusselCole - merging into upcoming release for QA.

PR #1029, opened 2021-09-14
PRESIDECMS-2203 object update filter skip useVersioning version record is not updated

pixl8-brayden:


DominicWatson:
Nice one, spotted a little typo in variable name, but fixing while merging. Merging into upcoming release for testing.

PR #1032, opened 2021-10-01
PRESIDECMS-2210 unique indexes with tenancy

chrisschmitz:
Fixes PRESIDECMS-2200. Unique indexes are not properly validated when used in objects with a tenant.

The reason he validation unique indexes previously failed was, that the tenant FK is not present in the form data and if one of the fields of a combined key is NULL, the validation is considered successful.

I added a new interception point 'preValidateForm' that calls the tenancy service, retrieves the tenant FK value and injects it into the form data before validation.

### DIFF
--- a/system/config/Config.cfc
+++ b/system/config/Config.cfc
@@ -183,6 +183,7 @@ component {
 		interceptorSettings.customInterceptionPoints.append( "postExtraTopRightButtonsForEditRecord" );
 		interceptorSettings.customInterceptionPoints.append( "postGetExtraCloneRecordActionButtons"  );
 		interceptorSettings.customInterceptionPoints.append( "postExtraTopRightButtons"              );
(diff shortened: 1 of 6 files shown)


DominicWatson:
Hi Chris, have finally had a chance to review this and generally looking really good. However, I think a couple of things:

1. The interception point announcement should be made in just one place (i.e. formsService.validateForm()) instead of twice in Datamanager.cfc - there will be a lot of other places in custom code that will miss this interception point.
2. If we are validating an update, we need to be getting the tenancy ID value from the current *record*, rather than from the request tenant ID

Point 2 might mean that we refactor the `getTenancyFieldsForInsertOrUpdateData`. Perhaps revert it to `getTenancyFieldsForInsertData` and add a new method for update data?

DominicWatson:
Fab, this is looking great - will have a couple of v minor tweaks to make but will keep you posted. Will merge into upcoming release and get some QA on this asap.

PR #1033, opened 2021-10-05
PRESIDECMS-1816 AJAX proxy XSS

teonator:


system/handlers/admin/AjaxProxy.cfc
```diff
@@ -12,8 +12,13 @@
 				var action = event.getValue( name="action", defaultValue="", statusCode=403 );
 				if ( Len( Trim( action ) ) ) {
 					event.noLayout();
+
 					try {
-						runEvent( "admin." & action );
+						if ( getController().handlerExists( "admin." & action ) ) {
```
DominicWatson, line 17: I'm not sure how this changes the behaviour regards to XSS. Logic inside `runEvent` will be running the same logic as `handlerExists()` and throw the same error.

I think the issue is more in the message that is output, i.e.
`"Action, [#action#], is not a valid handler action"`

I think simply a case of escaping `#action#` here??

DominicWatson (changes requested):

system/handlers/admin/AjaxProxy.cfc
teonator: `runEvent()` does not throw error when a handler does not exist. the issue is on the cfrequest include data.

PR #1035, opened 2021-10-06
PRESIDECMS-2212 XSS issues

jtcy0123:


system/services/rendering/LabelRendererService.cfc
```diff
@@ -61,7 +61,8 @@ component {
 				, eventArguments = args
 			);
 		} else {
-			return args.label ?: args[ $getPresideObjectService().getLabelField( objectname=args.object )] ?: "";
+
+			return args.label ? EncodeForHTML( args.label ) : args[ $getPresideObjectService().getLabelField( objectname=args.object )] ?: "";
```
DominicWatson: `args.label` is not a boolean. Also I think we should be encoding both when we have `label` and also when we fetch from a non default label field? So:

```cfc
if ( StructKeyExists( args, "label" ) ) {
    return EncodeForHtml( args.label );
}
return EncodeForHtml( args[ $getPresideObjectService().getLabelField( objectname=args.object ) ] ?: "" );
```

Or just wrap the original line in `EncodeForHtml`:

```cfc
return EncodeForHtml( args.label ?: ( args[ $getPresideObjectService().getLabelField( objectname=args.object ) ] ?: "" ) );
```

DominicWatson (changes requested):

DominicWatson:
Fab, will get merged and tested

DominicWatson:
We're going to reject this for now due to problems that are not easily fixable with this approach (i.e. select controls etc can't work with this html formatted labels). The source of the issue will need looking at in a different way.

PR #1037, opened 2021-10-11
PRESIDECMS-2117 2FA reset

teonator:


system/services/admin/LoginService.cfc
```diff
@@ -732,10 +732,8 @@ component displayName="Admin login service" {
 	 * @autodoc
 	 *
 	 */
-	public void function disableTwoFactorAuthenticationForUser() {
-		var userId = getLoggedInUserId();
-
-		if ( userId.len() ) {
+	public void function disableTwoFactorAuthenticationForUser( required string userId=getLoggedInUserId() ) {
```
DominicWatson: If there is a default, do not make the argument required. so should be `disableTwoFactorAuthenticationForUser( string userId=getLoggedInUserId() ) {...`

system/handlers/admin/EditProfile.cfc
```diff
@@ -209,6 +210,40 @@ component output="false" extends="preside.system.base.AdminHandler" {
 		setNextEvent( url = event.buildAdminLink( linkTo="editProfile.twoFactorAuthentication" ) );
 	}
 
+	function resetTwoFactorAuthenticationAction( event, rc, prc ) {
```
DominicWatson: This should be under `/handlers/admin/UserManager.cfc` (`EditProfile` is for editing your _own_ profile). And there should be a permission check also (same one as you have in the view, below).

DominicWatson (changes requested):
Looking great @teonator - couple of minor bits of feedback before merging.

PR #1038, opened 2021-10-12
PRESIDECMS-2214 batch all matching current filter

DominicWatson:
Adding to the work done to make batch edit and delete work in a background task - we have "select all records matching current filter". Merging this back into the background task branch as this is merged from there.


system/services/admin/DataManagerBatchOperationService.cfc
```diff
@@ -0,0 +1,429 @@
...
+			uriRoot = pobjService.getResourceBundleUriRoot( arguments.objectName );
+			objectTitle = $translateResource( uri=uriRoot & "title.singular", defaultValue=arguments.objectName );
+			fieldTitle  = $translateResource( uri=uriRoot & "field.#arguments.fieldName#.title", defaultValue=arguments.fieldName );
+			arguments.logger.info( $translateResource( uri="cms:datamanager.batchedit.task.starting.message", data=[ objectTitle, fieldTitle, NumberFormat( totalRecords ) ] ) );
+		}
+
+		try {
+			do {
+				if ( arguments.batchAll ) {
+					sourceIds = getNextBatchRecordsFromQueue( queueId, 100 );
```
Pixl8RusselCole: Slightly concerns me having 100 hard-coded in 2 places. Potentially someone could change the top one to 1000 without knowing to change the other

system/services/admin/DataManagerBatchOperationService.cfc
```diff
@@ -0,0 +1,429 @@
...
+		pobjService.insertDataFromSelect(
+			  objectName     = "batch_operation_queue"
+			, fieldList      = [ "queue_id", "record_id", "datecreated" ]
+			, selectDataArgs = queueDataArgs
+		);
+
+		return queueId;
+	}
+
+	public numeric function clearBatchOperationQueue( required string queueId ) {
```
Pixl8RusselCole, line 380: I wonder if there should be a message here to say how many records were removed from queue?

system/services/admin/DataManagerBatchOperationService.cfc
```diff
@@ -0,0 +1,429 @@
...
+			}
+
+			rethrow;
+		}
+
+		if ( Len( queueId ) ) {
+			clearBatchOperationQueue( queueId );
+		}
+		if ( canInfo ) {
+			arguments.logger.info( $translateResource( uri="cms:datamanager.batchedit.task.finished.message", data=[ objectTitle, fieldTitle, NumberFormat( totalRecords ) ] ) );
```
Pixl8RusselCole: Is totalRecords right in the finished.message? The task could have been interrupted before that number of records has been processed

PR #1040, opened 2021-10-13
PRESIDECMS-2027 add default image alt text

karhong:


system/forms/preside-objects/asset/group.upload.xml
```diff
@@ -11,6 +11,7 @@ that have been uploaded.
 		<fieldset id="standard" sortorder="10">
 			<field sortorder="10" binding="asset.asset_folder" control="assetFolderPicker" quickAdd="true" />
 			<field sortorder="20" binding="asset.author" control="textinput" />
+			<field sortorder="30" binding="asset.alt_text" control="textinput" />
```
DominicWatson: I'm not sure this makes sense here as this will set the same alt text for all images uploaded in a batch upload process. It should either just not be here, or added to the mini form you get for each asset (currently just allows you to edit the label).

system/i18n/preside-objects/asset_de.properties
```diff
@@ -22,6 +22,7 @@ field.full_login_required.title=Vollst\u00E4ndiger Login notwendig
 field.full_login_required.help=Ist ein Benutzer \u00FCber das "angemeldet bleiben"-Cookie angemeldet, wird er nicht als vollst\u00E4ndig angemeldet behandelt. 
 field.focal_point.title=Brennpunkt
 field.crop_hint.title=Zuschneidebereich
+field.alt_text.title=Alternativer Text
```
DominicWatson: Translations need to come from Zanata. This change should be removed.

DominicWatson (changes requested):
Looking great - a couple of minor points for review but will be great to get this in for 10.16

DominicWatson:
Great stuff - merging into upcoming release

PR #1042, opened 2021-10-29
PRESIDE-2220 asset browserjs update

annbelievable:
This is to enable a way to override the default number of items displayed in the asset manager


DominicWatson:
Thanks for this Kok Ann. I'm not a big fan of this solution though. The requirement is:

> AS a developer, I want the flexibility to set custom pagination for datatables, in order to satisfy different client scenarios and preferences

^^ potentially even AS a user.

I think this PR kind of achieves that, but I think we should aim for a more unified approach that can be documented. I'm going to knock something up now.

PR #1043, opened 2021-11-11
PRESIDECMS-2226 `select-data-view` properties should not have db field

choontat-pixl8:
Do not create DB field and rules engine expressions for `select-data-view` properties


DominicWatson:
Nice one, thanks @choontat-pixl8 - merging into bleeding edge for some QA before likely hotfixing.

PR #1044, opened 2021-11-12
PRESIDECMS-2227 Error Exporting Formbuilder v2 Submission that has Checkboxlist field with Values only no Labels

jtcy0123:


DominicWatson:
Nice one, thanks @jtcy0123  - will get merged and tested and bundle into a hotfix soon.

PR #1049, opened 2021-11-26
PRESIDECMS-2038 make sure the needed argument value is passed along

annbelievable:
This is to fix the 404 error when admin tries to filter the stats in system email template. This fix the error by making sure the needed argument value is passed along


annbelievable:
just realised i used the wrong issue number, will create a new branch and new MR.

PR #1052, opened 2021-11-30
PRESIDECMS-2236 Delete prompt empty label

teonator:


DominicWatson:
Nice one, looks like this one got dropped. Getting merged now for QA.

PR #1053, opened 2021-12-07
Feature presidecms 2228 data exported for enum field are enum value instead of enum label

Pixl8RusselCole:


DominicWatson:
Looking good, thanks @Pixl8RusselCole . Looks like some odd indentation in this file, but I'll fix that.

PR #1056, opened 2021-12-13
PRESIDECMS-2247 formbuilder content cannot be updated when more than one content or spacer field

Pixl8RusselCole:


DominicWatson:
Nice, thanks @Pixl8RusselCole - looks good. Will merge into upcoming release for QA.

PR #1060, opened 2021-12-29
PRESIDECMS-2255 Extending forms can sometimes throw 500 error

choontat-pixl8:
Do not modify the original extended form definition when resolving extensions


DominicWatson:
Nice one, thanks for this @choontat-pixl8! Merging into the upcoming release for QA before release.

PR #1064, opened 2022-01-10
PRESIDECMS-2260 Formbuilder v1 response not render after v2 enabled

teonator:


system/handlers/Tasks.cfc
```diff
@@ -103,4 +104,16 @@ component {
 	private boolean function deleteExpiredQueuedAssetGenerations( logger ) {
 		return assetQueueService.deleteExpiredQueuedItems( arguments.logger ?: NullValue() );
 	}
+
```
DominicWatson, line 106: I'm not sure I like the idea of a non-scheduled task that appears in task manager by default with Preside.

Should/could this be a one off, onApplicationStart() type routine? It should not run if it has run already - i.e. check and set a preside system setting. Then run it once on startup if needed?

DominicWatson (changes requested):
Looking good @teonator - I just have a question around the task manager task..

system/handlers/Tasks.cfc
teonator: Changed to use one-off update using onApplicationStart().

PR #1066, opened 2022-01-12
PRESIDECMS-2262 auto redirect check error when sites feature disabled

pixl8-brayden:
Add structe key exist check before checking on the value


DominicWatson:
Nice one, thanks @pixl8-brayden . I'm going to pull this but will tweak to make more concise. i.e.

```cfc
return isBoolean( site.auto_redirect ?: "" ) && site.auto_redirect;
```

PR #1067, opened 2022-01-17
Feature presidecms 2264 contentrendererservice renderlabel render id for no label objects

Pixl8RusselCole:


system/services/rendering/ContentRendererService.cfc
```diff
@@ -62,6 +62,10 @@ component {
 		,          string labelRenderer = $getPresideObjectService().getObjectAttribute( arguments.objectName, "labelRenderer" )
 		,          array bypassTenants = []
 	) {
+		var noLabel = $getPresideObjectService().getObjectAttribute( arguments.objectName, "noLabel", "" );
```
DominicWatson: I'm not sure this is quite right. An object can define both `@nolabel` and also specify a `@labelRenderer`. We should do our best to render something and fall back to returning ID only when we can't.

DominicWatson (changes requested):

DominicWatson:
Nice one, merging into upcoming release for QA.

PR #1069, opened 2022-01-26
PRESIDECMS-2267 object translation error when creating transations contain specific record condition rule

pixl8-brayden:
for case when `obj` returned as simple value (i.e auto_created) then skip the `getDateModifiedField()` latest value for cache buster.


DominicWatson:
Nice one - thanks for digging this out @pixl8-brayden . Will merge in upcoming for QA.

PR #1075, opened 2022-02-23
PRESIDECMS-2279 pass changed data as parameter to update data

Pixl8RusselCole:


system/services/presideObjects/PresideObjectService.cfc
```diff
@@ -691,43 +692,44 @@ component displayName="Preside Object Service" {
 				);
 			}
 
-			arguments.changedData = {};
-			var versionedManyToManyFields = _getVersioningService().getVersionedManyToManyFieldsForObject( arguments.objectName );
-			for( var record in arguments.oldData ) {
-				var oldManyToManyData = {};
+			if ( structIsEmpty( arguments.changedData ?: {} ) ) {
```
sebduggan: Don't need the fallback to an empty struct. The argument itself has a default value, so it will always be a struct...

sebduggan (changes requested):
One small comment...

sebduggan (approved):

PR #1076, opened 2022-02-25
Feature presidecms 2266 generate preview link

john5onCheng:
Could we allow preview to generate view online link? We have request from user who want to share the link manually to member


DominicWatson:
Nice one, thanks @john5onCheng - merging into upcoming release for QA

PR #1077, opened 2022-02-25
PRESIDECMS-2282: Multilingual auto-joins with partly drafts enabled result in error

jjannek:
If multilingual objects are auto-joined and the root object has drafts enabled but the object to be joined not, an error occurs because a db column is used which does not exist: _version_is_latest_draft

This is caused by the fact that Preside only checks the root main object if drafts are enabled or not. Target tables of joins are not checked.
So if there is on object which allows versioning, translations and drafts - and this joins to another object which has has versioning and translations enabled but not drafts (which is a pretty common scenario), then this results in an error.

The solution is to also check if the target table of a join belongs to an object that has drafts enabled.
Therefore I created a small lookup cache in the PresideObjectService that let's you find the Preside Object Name by a given table name.
Within the interception to add the translation checks for latest versions this is now used to determine whether the target table has drafts enabled or not.

One small additional off topic fix:
In Preside Datamanager the top right buttons in the View-Record screen are misaligned visually if translations are enabled.
If fixed this by changing the order of the buttons, putting the translation dropdown first in line. This is now also aligned to the edit record screen, where this was already the case.

### DIFF
--- a/system/handlers/admin/DataManager.cfc
+++ b/system/handlers/admin/DataManager.cfc
@@ -1655,12 +1655,29 @@ component extends="preside.system.base.AdminHandler" {
 	}
 
 	private array function getTopRightButtonsForViewRecord() {
-		var objectName  = args.objectName ?: "";
+		var objectName   = args.objectName ?: "";
 		var objectTitle  = prc.objectTitle ?: "";
 		var recordId     = prc.recordId    ?: "";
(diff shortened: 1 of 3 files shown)


DominicWatson:
Great stuff @jjannek - I'm going to merge this into upcoming release branch. If you could QA the issue(s) fixed here in your environment we can get this released in the next hotfix.

PR #1079, opened 2022-02-28
Feature presidecms 2283 formbuilder v2 form not sending post action to anonymous user

chunwengpixl8:
Fixing bug in FormBuilderService in order to get the v2 responses in saveFormSubmission and updated AnonymousCustomerEmail handler so that the v2 responses will be picked up.


DominicWatson:
Looking good, thanks @chunwengpixl8 . Merging into upcoming release for QA

PR #1081, opened 2022-03-03
PRESIDECMS-2286 Password error msg appear duplicated

jtcy0123:
Updated password score checker to hide field error div if there is any before showing password score & policies


DominicWatson:
Looks good thanks @jtcy0123 ! Will merge into upcoming release for QA

PR #1082, opened 2022-03-04
PRESIDECMS-2284 Admin view record property no value

teonator:


system/handlers/admin/DataHelpers.cfc
```diff
@@ -153,15 +158,19 @@ component extends="preside.system.base.adminHandler" {
 		var baseLink      = event.buildadminLink( objectName=relatedObject, recordId="{recordId}" );
 		var list          = [];
 
-		for( var record in records ) {
-			if ( baseLink.len() ) {
-				list.append( '<a href="#( baseLink.replace( '{recordId}', record.id ))#">#record.label#</a>' );
+		for ( var record in records ) {
+			var recordLabel = record.label;
+
+			recordLabel = _renderyNoValue( objectName=objectName, propertyName=propertyName, propertyValue=recordLabel );
```
DominicWatson: Am not sure why `_renderyNoValue()` is required here? - we know that we have a list of records, so there should always be something to render?

system/handlers/admin/DataHelpers.cfc
```diff
@@ -281,4 +290,21 @@ component extends="preside.system.base.adminHandler" {
 		}
 	}
 
+	private string function _renderyNoValue(
```
DominicWatson: Should this be `_renderNoValue()` rather than `_renderyNoValue()`?

DominicWatson (changes requested):
Nice one, looking good @teonator . Could you check the review comments and update?

system/handlers/admin/DataHelpers.cfc
teonator: Removed.

system/handlers/admin/DataHelpers.cfc
teonator: Fixed.

DominicWatson:
Fab, merging into upcoming release for QA

PR #1083, opened 2022-03-04
EmailStyleInliner fix

jjannek:
See https://presidecms.atlassian.net/browse/PRESIDECMS-2288

There is a problem with HTML snippets consisting of only TD or TR wrapped content, without needed surrounding table tags. Internal JSOUP parsing will strip out HTML markup and email content ends up not being what was expected.

The fix adds dummy table and tr tags and removes them again after JSOUP parsing.

### DIFF
--- a/system/services/email/EmailStyleInliner.cfc
+++ b/system/services/email/EmailStyleInliner.cfc
@@ -45,8 +45,24 @@ component {
 			return fromCache;
 		}
 
+		arguments.html = trim( arguments.html );
+
 		var innerHtmlOnly = !FindNoCase( "</html>", arguments.html );
-		var doc           = _jsoup.parse( arguments.html );
+
+		// special cases of widget which only consist of a table cell or row without a wrapping table tag
+		// jsoup will remove the TD / TR tags in those cases, therefore adding now and stripping after processing
+		var isHtmlTableCell = innerHtmlOnly && ReFindNoCase( "^<td[^>]*>", arguments.html );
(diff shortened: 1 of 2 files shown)


system/services/email/EmailStyleInliner.cfc
```diff
@@ -45,8 +45,24 @@ component {
...
 		var innerHtmlOnly = !FindNoCase( "</html>", arguments.html );
-		var doc           = _jsoup.parse( arguments.html );
+
+		// special cases of widget which only consist of a table cell or row without a wrapping table tag
+		// jsoup will remove the TD / TR tags in those cases, therefore adding now and stripping after processing
+		var isHtmlTableCell = innerHtmlOnly && ReFindNoCase( "^<td[^>]*>", arguments.html );
+		var isHtmlTableRow  = innerHtmlOnly && !isHtmlTableCell && ReFindNoCase( "^<tr[^>]*>", arguments.html );
+
+		// add dummy wrapping html table and row tags to make sure jsoup parsing works as expected
+		if ( isHtmlTableCell ) {
```
DominicWatson, line 58: I suggest adding an id to the first containing element in both of these, i.e.

```cfc
arguments.html = "<table><tbody id='_css-inliner-wrapper'> & ...
```

Then, if using this, you can get the inner html of this element from the `doc`, rather than doing regex at the end.

DominicWatson (changes requested):

DominicWatson:
Nice one, merging for QA

PR #1084, opened 2022-03-04
Feature presidecms 2289 conditional content render for json

Pixl8RusselCole:


DominicWatson:
This is not quite the right approach to fix the issue. Closing. (discussion had outside of this PR)

PR #1085, opened 2022-03-15
PRESIDECMS-2294 make sure sync many to many field doesnt update sort order when sortable set to false

karhong:


DominicWatson:
This feels a little janky to me. The `sortable=false` that you have on your property definition is a form control instruction - so using it as something in the data model like this feels less than ideal.

I think that the source issue is that for auto pivot tables, we _always_ create a sort order column. If this were configurable in the property, I think it would solve the problem without any further code changes.

The logic for creating the pivot table definition can be found in `PresideObjectReader.cfc$getAutoPivotObjectDefinition()`.

We could detect the `sortable` attribute on the property and _not_ create the sort_order column on the pivot table when it is false:

```
property name="categories" relationship="many-to-many" relatedto="category" relatedvia="myobj_category" sortable=false;
```

Could you try this approach to see if it solves the problem?

karhong:
Yes, I do agree the  `sortable=false` might not a good way to do this. But the issue here isn't to remove sort_order column, since it is still needed to sort the items, but not sort the categories. Because we only have one sort_order column in pivot table which shared by two objects, which isn't ideal...To resolve this, I think either we should only allow sort_order to be used on one object only, or we can have two sort_order columns, in this case E.g `item_sort_order` and `category_sort_order`

PR #1089, opened 2022-03-22
PRESIDECMS-2298 Formcontrol/ Formbuilder url field

teonator:


system/assets/js/admin/specific/urlInput/urlInput.js
```diff
@@ -0,0 +1,26 @@
+( function( $ ) {
```
DominicWatson, line 1: It would be good/essential to add in logic to pick up a protocol pasted into the address field and automatically remove it and set the protocol input based on it. i.e. if user pastes `http://www.myoldsite.com`, the control will automatically change it to `www.myoldsite.com` and set the protocol field to `http://`

system/forms/formbuilder/item-types/url.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<form i18nBaseUri="formbuilder.item-types.url:">
+	<tab id="default">
+		<fieldset id="default">
+			<field name="protocol" control="select" values="http://,https://" labels="http://,https://" defaultvalue="" />
```
DominicWatson: This feels a little tough to explain for admin users. I'd rather we didn't make this a configurable option at this point to keep the system simpler. i.e. limiting user input to a specific protocol isn't all that helpful(?)

system/views/formcontrols/urlInput/index.cfm
```diff
@@ -0,0 +1,53 @@
...
+	value = event.getValue( name=inputName, defaultValue=defaultValue );
+	if ( !IsSimpleValue( value ) ) {
+		value = "";
+	} else if ( REFindNoCase( "^https?:\/\/([-_A-Z0-9]+\.)+[-_A-Z0-9]+(\/.*)?$", value ) ) {
+		protocolDefaultValue = ArrayFirst( REMatch( "^https?:\/\/", value ) );
+		domainPathDefaultValue = ReplaceNoCase( value, protocolDefaultValue, "" );
+	}
+
+	protocol       = args.protocol ?: "";
+	protocolValues = isEmptyString( protocol ) ? "http://,https://" : protocol;
```
DominicWatson: `https://` should be first in the list and the default value selected.

DominicWatson (changes requested):

system/assets/js/admin/specific/urlInput/urlInput.js
```diff
@@ -2,17 +2,20 @@
...
+				var $hidden     = $( this ).closest( ".row" ).next( 'input[type="hidden"]' )
+				  , inputId     = $hidden.attr( "id" )
 				  , $protocol   = $( "#" + inputId + "_protocol" )
 				  , $domainPath = $( "#" + inputId + "_domain_path" )
 				;
 
 				if ( $protocol.val().length > 0 && $domainPath.val().length > 0 ) {
+					var matches = $domainPath.val().match( /^(.*:\/\/)/g );
+
+					$protocol.val( matches ? matches[ 0 ] : "https://" );
```
DominicWatson: Are we saying that if someone enters a URL _without_ a protocol, we always set the protocol to https whenever a change is detected?

I think the value should only get set if:

1. Protocol is found in the address
2. Or, the protocol is currently empty

DominicWatson (changes requested):

system/assets/js/admin/specific/urlInput/urlInput.js
```diff
@@ -2,17 +2,29 @@
...
+			;
+
+			$( '[name="' + inputName + '_protocol"],#' + inputId + "_domain_path").on( "change", function( e ) {
+				var $protocol   = $( '[name="' + inputName + '_protocol"]' )
 				  , $domainPath = $( "#" + inputId + "_domain_path" )
 				;
 
-				if ( $protocol.val().length > 0 && $domainPath.val().length > 0 ) {
-					var matches = $domainPath.val().match( /^(.*:\/\/)/g );
+				if ( $domainPath.val().length > 0 ) {
```
DominicWatson: This still doesn't look right?

* If a protocol is detected in domain path then, **always set the protocol** using that value
* If a protocol is _not_ detected in the domain path and protocol field is empty, then set the protocol field to https

DominicWatson (changes requested):

DominicWatson:
Nice one,  merging into upcoming release for QA

PR #1090, opened 2022-03-23
PRESIDECMS-2299 allow admin to use and manage filter for rules engine condition

choontat-pixl8:


DominicWatson:
Great, thanks @choontat-pixl8 - merging into upcoming release for QA, etc.

PR #1091, opened 2022-03-24
PRESIDECMS-2300 Crontab definition validation for saved exports

choontat-pixl8:
Add the same validation to the saved export crontab definition field that are applied on task manager crontab definition field


DominicWatson:
Great, thanks @choontat-pixl8 ! Will get merged into upcoming release for pre-release QA.

PR #1097, opened 2022-04-11
PRESIDECMS-2305 tenancy extra filter objectname is not versioned

Pixl8RusselCole:


DominicWatson:
Nice one, merging for QA

PR #1101, opened 2022-04-15
PRESIDECMS-2310 Render enum

teonator:
Create a simple render enum helper.

### DIFF
--- a/system/services/rendering/ContentRendererService.cfc
+++ b/system/services/rendering/ContentRendererService.cfc
@@ -133,19 +133,25 @@ component {
 	}
 
 	public string function renderEnum(
-		  required string objectName
-		, required string propertyName
-		, required string data
+		  required string data
+		,          string enum         = ""
+		,          string objectName   = ""
+		,          string propertyName = ""
 		,          any    context      = "default"
 		,          string recordId     = ""
-		,          string enumRenderer = $getPresideObjectService().getObjectAttribute( arguments.objectName, "enumRenderer", "enumLabel" )
+		,          string enumRenderer = "enumLabel"
 	) {
+		if ( !$helpers.isEmptyString( arguments.objectName ) ) {
+			arguments.enumRenderer = $getPresideObjectService().getObjectAttribute( arguments.objectName, "enumRenderer", "enumLabel" );
+		}
+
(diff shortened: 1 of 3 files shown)


DominicWatson:
This is cool as is. However, have also been looking at this recently and would make sense to add one general improvement:

1. Allow passing enum directly, rather than forcing `objectName` and `propertyName` combo to get the enum from preside object layer 

So:

```cfc
// new 
#renderEnum( enum="orderStatus", data="pending" )#

// and still support
#renderEnum( objectName="order", propertyName="status", data=record.status )#
```

system/services/rendering/ContentRendererService.cfc
```diff
@@ -133,19 +133,25 @@ component {
...
+		,          string enum         = ""
+		,          string objectName   = ""
+		,          string propertyName = ""
 		,          any    context      = "default"
 		,          string recordId     = ""
-		,          string enumRenderer = $getPresideObjectService().getObjectAttribute( arguments.objectName, "enumRenderer", "enumLabel" )
+		,          string enumRenderer = "enumLabel"
 	) {
+		if ( !$helpers.isEmptyString( arguments.objectName ) ) {
+			arguments.enumRenderer = $getPresideObjectService().getObjectAttribute( arguments.objectName, "enumRenderer", "enumLabel" );
```
DominicWatson: Previously to this commit, you were able to specify an `enumRenderer` that was different to what was in the object property definition.

Perhaps the default enumRenderer arguments value should be `""`. Then use `getObjectAttribute()` when objectName supplied and just "enumLabel" otherwise?

Also, should it really be the _object_ attribute and not the _property_ attribute??

DominicWatson (changes requested):
Nice one - tiny bit of feedback but this looks good otherwise

system/services/rendering/ContentRendererService.cfc
teonator: Updated to default enumRenderer value to `""` and replaced `getObjectAttribute()` with `getObjectPropertyAttribute()`.

DominicWatson:
Great stuff. Getting this merged for 10.18 testing.

PR #1107, opened 2022-04-26
PRESIDECMS-2323 Datepicker on render

teonator:
Add datepicker onRender event.


system/assets/js/admin/lib/plugins/bootstrap.datepicker.js
```diff
@@ -430,7 +431,7 @@
 
 					}
 				}
-				clsName = '';
+				clsName = ' '+this.onRender(prevMonth)+' ';
```
DominicWatson, line 434: Its unclear to me what the purpose is here and how it would be used. Are we saying that an `onRender` customisation of this datepicker would return a class name that is expected to be added to the wrapper?

Could you provide some examples of how this would be used and why?

DominicWatson:
Closing for now.

PR #1111, opened 2022-05-19
PRESIDECMS-2333 multi select panel form control out of index error

pixl8-brayden:
Check the values length and if greater than it then fall to the length of the values


DominicWatson:
Nice one, merging for QA.

PR #1116, opened 2022-06-07
Feature presidecms 2339 configure preside so that reload all fwreinit true can be disabled

Pixl8RusselCole:


system/handlers/admin/devtools/terminalCommands/Reload.cfc
```diff
@@ -32,13 +34,21 @@ component hint="Reload all or part of your preside application" {
...
 		if ( environment == "production" && ( target.flagRequiredInProduction ?: false ) && !forceFlag ) {
 			return Chr(10) & "[[b;red;]--force flag is required to perform this action in a production environment]" & Chr(10);
 		}
 
+		var allowReloadTarget = IsBoolean( allowReload[ targetName ] ?: "" ) ? allowReload[ targetName ] : true;
+		if ( !allowReloadTarget ) {
+			return Chr(10) & "[[b;red;]This action has been disallowed]" & Chr(10);
+		}
+
+return Chr(10) & "[[b;red;]Action would be allowed]" & Chr(10);
```
DominicWatson: Debug left in place.

system/config/Config.cfc
```diff
@@ -95,6 +95,13 @@ component {
 	private void function __setupColdbox() {
 		var applicationSettings = getApplicationSettings();
 
+		settings.allowReload    = {
```
DominicWatson: I feel like we should/could be more brutal here to simplify things. Just:

```cfc
settings.disableMajorReloads = IsBoolean( settings.env.DISABLE_MAJOR_RELOADS ?: "" ) && settings.env.DISABLE_MAJOR_RELOADS;

// ...
// ...
, reinitPassword = settings.disableMajorReloads ? Hash( CreateUUId() ) : ( applicationSettings.COLDBOX_RELOAD_PASSWORD ?: "true" )
```
^^ This effectively prevents `?fwreinit=true` in the URL. We can also then use this single setting in General.cfc and reload devtools cfc to bypass checking when the reload type is a major one (with less code).

I think the ability to have this fine tuned is not really necessary.

DominicWatson (changes requested):
Good start. I've left some review feedback that I think will give us complete coverage and simplify the changes somewhat.

DominicWatson:
:boom: looking good. Merging into upcoming release for QA.

PR #1119, opened 2022-06-14
PRESIDECMS-2342 form with required fields pass validation with empty string

karhong:


DominicWatson:
Nice one, thanks @karhong - amazed this has never been picked up.

PR #1128, opened 2022-07-12
PRESIDECMS-2346 site tree pages show blank for admin when no version records

karhong:


DominicWatson:
Yes, this. Excellent. Let's get through QA.

PR #1134, opened 2022-07-19
PRESIDECMS-2356 Formbuilder submission files deletion

teonator:


system/interceptors/FormBuilderInterceptor.cfc
```diff
@@ -6,4 +6,25 @@ component extends="coldbox.system.Interceptor" {
...
+
+			if ( isArray( id ) ) {
+				id = ArrayFirst( id );
+			}
+
+			if ( !isEmptyString( id ) ) {
+				formBuilderService.deleteSubmissionResponses( submissionId=id );
+			}
+
+			if ( !isEmptyString( filter.form ?: "" ) ) {
```
DominicWatson: This feels potentially dangerous. *IF* for example, some code ran that did:

```cfc
deleteData( filter={ form=formId, user=userid } );
```

^^ we'd end up deleting entries for the entire form.

Perhaps this?

```cfc
if ( !isEmptyString( filter.form ?: "" ) && StructCount( filter ) == 1 ) {
     // ...?
}
```

Or another approach entirely that should allow most flexibility and coverage:

```cfc
if ( !isEmptyString( id ) ) {
    ...
    return;
}

formBuilderService.deleteFormResponses( responseFilter=filter ); << then use that filter to delete them
```

system/services/formbuilder/FormBuilderService.cfc
```diff
@@ -1141,6 +1142,104 @@ component {
...
 
+	/**
+	 * Delete the given submission files.
+	 *
+	 * @autodoc
+	 * @submissionId.hint The ID of the submission
+	 *
+	 */
+	public void function deleteSubmissionFiles( required string submissionId ) {
+		var submission    = getSubmission( submissionId=arguments.submissionId );
```
DominicWatson, line 1153: I wonder if there is anyway to background task this _after_ the form / response has been deleted. If we delete a large form with a lot of entries, this process could take a long while and create a very slow request.

Easiest way perhaps would be to make the delete form flow use an adhoc bg task with a progress UI for the user to view it. Same for submission. That way, the task can take a while and the user can navigate away, etc.

DominicWatson (changes requested):
Nice one, looking good on the whole. Have a bit of feedback for potential trouble.

system/services/formbuilder/FormBuilderService.cfc
```diff
@@ -1141,6 +1142,104 @@ component {
...
+			}
+		}
+
+		if ( ArrayLen( files ) ) {
+			for ( var file in files ) {
+				if ( !$helpers.isEmptyString( file ) ) {
+					var filePaths = ListToArray( file );
+
+					for ( var filePath in filePaths ) {
+						formBuilderStorageProvider.deleteObject( path=filePath, private=formBuilderStorageProvider.objectExists( path=filePath, private=true ) );
```
DominicWatson, line 1198: We ought to try/catch here. If a file no longer exists, for example (or some other remote file system error), it shouldn't prevent the entire form from being deleted.

```cfc
try {
 ...
} catch( any e ) {
    $raiseError( e );
}
```

DominicWatson:
Nice one, a bit of feedback, but I'll action that.

PR #1136, opened 2022-07-22
Feature presidecms 2363 rules field interceptor

mookeshNanda:
Added pre/post interceptor point for getRulesForField()


DominicWatson:
Nice one, thanks @mookeshNanda . Will be useful. Will get merged for QA.

PR #1140, opened 2022-08-02
PRESIDECMS-2366 length check missing for text dbtype field and cause error

pixl8-brayden:


DominicWatson:
Nice one, merging for QA.

PR #1144, opened 2022-08-05
PRESIDECMS-2369 greater than and lesser than core validators

pixl8-brayden:


DominicWatson (changes requested):
Looks great, but:

* can rename to just `lessThanField` (rather than "lesser")
* makes sense to also add `lessThanOrEqualToField` + `greaterThanOrEqualToField` validators

pixl8-brayden:
Updated and added 2 equal to validators, could you have another look? thanks

PR #1147, opened 2022-08-16
PRESIDECMS-1816 Ajax proxy xss cfrequest data

teonator:


system/base/AdminHandler.cfc
```diff
@@ -25,7 +25,12 @@ component {
 			, adminBaseUrl = event.getAdminPath()
 			, siteId       = event.getSiteId()
 		} );
-		event.includeData( event.getCollection() );
+
+		var data = event.getCollection();
+		for ( var key in data ) {
+			data[ key ] = HtmlEditFormat( data[ key ] );
```
DominicWatson: I'm not sure that we can just escape the data this way. This will transform the data and presumably break any legitimate data that contains any characters that would be escaped. I think an alternative approach will be to force it through AntiSamy:

```cfc
property name="antiSamySettings" inject="coldbox:setting:antiSamy";
property name="antiSamyService"  inject="antiSamyService";

// ...

// only necessary if antisamy is disabled or disabled in admin
if ( IsFalse( antiSamySettings.enabled ?: "" ) || IsTrue( antiSamySettings.bypassForAdministrators ?: "" ) ) {
	var policy = antiSamySettings.policy ?: "myspace";
	var data   = StructCopy( event.getCollection() );

	for( var key in data ) {
		data[ key ] = antiSamyService.clean( data[ key ], policy );
	}
}
```

DominicWatson (changes requested):

DominicWatson:
Nice one, merged into upcoming release for qa

PR #1149, opened 2022-08-30
PRESIDECMS-2371 batch update sometime get all records instead of filtered

karhong:


DominicWatson:
Nice one, merging into upcoming release for QA.

PR #1150, opened 2022-09-01
PRESIDECMS-2379 Render label with numeric key field

teonator:


DominicWatson:
Fab, getting merged for QA.

PR #1154, opened 2022-09-13
PRESIDECMS-2384 Condition picker filter condition filter object context indicator

teonator:


DominicWatson:
Love it. merging into 10.19 upcoming release

PR #1156, opened 2022-09-22
PRESIDECMS-2391 multiformbuilder on a page issue

mookeshNanda:
Add formId to be part of the idPrefix hash to ensure each forms and inputs will have unique id when multiple form used on same page


DominicWatson:
Fab, great solution @mookeshNanda . Will get merged and ready for QA.

PR #1158, opened 2022-09-30
Feature presidecms 2392 include contenteditable in useristyping condition to prevent triggering shortcut keys whilst typing

nickcervantes:
- Include 'contenteditable is true' condition to determine if `userIsTyping`. This will prevent triggering preside schortcut keys while typing, ie. 'e' for quick edit mode.


DominicWatson:
Nice one, thanks @nickcervantes . Looks good. Will get merged and setup a test for QA.

PR #1159, opened 2022-10-03
PRESIDECMS-2393 saved export failed to generate report when currently is running

pixl8-brayden:


DominicWatson:
Nice one, thanks @brayden. Will get merged for testing.

PR #1171, opened 2022-10-17
PRESIDECMS-2403 admin users reset system email templates intermediate screen

pixl8-brayden:


system/helpers/presideProxies.cfm
```diff
@@ -300,6 +300,15 @@
 		</cfscript>
 	</cffunction>
 
+<!--- system email --->
+	<cffunction name="systemEmailBodyIsDiffWithDefault" access="public" returntype="string" output="false">
```
DominicWatson: This really doesn't need/want to be a helper function. No one will use this. In the case where it is being used, a variable should be set in the handler, i.e.

```cfc
prc.templateBodyIsDifferentFromDefault = systemEmailTemplateService.bodyIsDifferentWithDefault( ... );
```

^^ then query that variable in the view.

DominicWatson (changes requested):

PR #1175, opened 2022-11-01
PRESIDECMS-2407 js boolean condition not working in useristyping

nickcervantes:
Removed `== true` with the .prop() in the condition.


DominicWatson:
Nice one, will get into next hotfix

PR #1177, opened 2022-11-04
PRESIDECMS-2409 foreign property calculation for formula property expressions

choontat-pixl8:
Make use of the calculated prefixes in the `having` clause


DominicWatson:
Great work figuring this out @choontat-pixl8 . Merging into upcoming release for QA.

PR #1178, opened 2022-11-08
PRESIDECMS-2399 Asset manager thumbnail

teonator:


teonator:
Replaced with https://github.com/pixl8/Preside-CMS/pull/1179

PR #1182, opened 2022-11-15
PRESIDECMS-2414 manytoone filter and having exist error during prepare filter

pixl8-brayden:


DominicWatson:
Great stuff. Simple change, tough to track down!

PR #1186, opened 2022-11-18
PRESIDECMS-2417 error in saving filter when no input optional filter

karhong:


DominicWatson:
Fab, merged into upcoming release for QA

PR #1190, opened 2022-11-25
Feature presidecms 2424 dbmigrations domigration check isenabled handler

Pixl8RusselCole:


system/services/updateManager/AppDatabaseMigrationService.cfc
```diff
@@ -26,6 +26,19 @@ component {
 
 		$SystemOutput( "Running DB Migration: [#arguments.migration#]..." );
 
+		if ( $getColdbox().handlerExists( "dbmigrations.#arguments.migration#.isEnabled" ) ) {
```
DominicWatson: This is logically in the wrong place I feel. This logic should be in the *_getInstalledMigrations()* call. i.e. if they have `isEnabled()` and it returns false, we should consider them not installed.

DominicWatson (changes requested):

DominicWatson:
Perfect. Merging for Qa.

PR #1191, opened 2022-11-30
PRESIDECMS-2425 Formbuilder post action fails to send email to anonymous user

teonator:


DominicWatson:
Nice one, merging for qA

PR #1193, opened 2022-12-05
PRESIDECMS-2419 system alerts

sebduggan:


system/handlers/renderers/content/SystemAlertReference.cfc
```diff
@@ -0,0 +1,9 @@
+component {
+
+	private string function default( event, rc, prc, args={} ){
+		var reference = args.data ?: "";
+
+		return Len( reference ) ? reference : #translateResource( "cms:preside-objects.default.field.no_value.title" )#;
```
DominicWatson: unneeded `#` around `translateResource()` here

system/interceptors/SystemAlertsInterceptor.cfc
```diff
@@ -0,0 +1,18 @@
+component extends="coldbox.system.Interceptor" {
+	property name="systemAlertsService" inject="delayedInjector:SystemAlertsService";
+
+	public void function configure() {}
+
+	public void function onApplicationStart() {
+		systemAlertsService.runStartupChecks();
```
DominicWatson: Could/should this be moved to General.cfc$applicationStart()?

Would be good not to need any interception points here in core given we know that we're going to run this code (so not an extension point really)?

system/interceptors/SystemAlertsInterceptor.cfc
```diff
@@ -0,0 +1,18 @@
...
+	}
+
+	public void function onClearSettingsCache( event, interceptData ) {
+		var category = interceptData.category ?: "";
+		if ( category == "dynamicform" ) {
```
DominicWatson: Could this go in the system config service where this interceptor is fired (after it) - rather than in interceptor?

system/services/systemAlerts/SystemAlertsService.cfc
```diff
@@ -0,0 +1,459 @@
...
+				  event             = "admin.systemAlerts.runCheckInBackgroundThread"
+				, args              = { type=arguments.type, reference=arguments.reference }
+				, runNow            = true
+				, discardOnComplete = false
+			);
+			return;
+		}
+
+		var checks     = [];
+		var references = config.isMultiCheck && !Len( arguments.reference ) ? $runEvent(
```
DominicWatson, line 42: This feels a little loose. i.e. if the `references` handler action does not exist, the error won't be all that instructional for a developer.

Should we be checking that the handler exists first and then throwing some explicit error when expecting references and we don't have them (either because not passed or no handler exists, or handler not returning correct/any data?

DominicWatson (changes requested):

PR #1195, opened 2022-12-07
PRESIDECMS-2429 one to many configurator form control stop working

karhong:


sebduggan:
Thanks @karhong 

However, there is a more targeted (and non-temporary) fix that checks if the related object is a configurator object, so will implement that instead.

PR #1197, opened 2022-12-14
PRESIDECMS-2430 adhoc task log history not appearing to auto cleanup based on settings

pixl8-brayden:


system/services/taskmanager/AdHocTaskManagerService.cfc
```diff
@@ -10,18 +10,21 @@ component displayName="Ad-hoc Task Manager Service" {
...
- 	 * @executor.inject    presideAdhocTaskManagerExecutor
+	 * @siteService.inject   siteService
+	 * @sysConfigSrvc.inject systemConfigurationService
+	 * @threadUtil.inject    threadUtil
+	 * @logger.inject        logbox:logger:taskmanager
+ 	 * @executor.inject      presideAdhocTaskManagerExecutor
 	 */
 	public any function init(
 		  required any siteService
+		, required any sysConfigSrvc
```
DominicWatson: I don't think we need to inject this service here. We can just use methods from the super class, i.e.

```cfc

var daysToKeepLogs = Val( $getPresideSetting( "taskmanager", "keep_logs_for_days", 7 ) );
```

tests/integration/api/taskmanager/AdHocTaskManagerServiceTest.cfc
```diff
@@ -796,6 +796,7 @@ component extends="testbox.system.BaseSpec" {
 		mockRequestContext  = CreateStub();
 		mockTaskScheduler   = CreateStub();
 		mockSiteService     = CreateStub();
+		mockSysConfigSrvc   = CreateStub();
```
DominicWatson: This test change not necessary with feedback above.

DominicWatson (changes requested):
Looking good, just a little comment about using core super class rather than injecting a new service.

DominicWatson:
Fab, merging for QA.

PR #1199, opened 2022-12-15
PRESIDECMS-2411 valid sending domain check can fail with deleted system emails

pixl8-brayden:


system/services/validation/EmailCenterValidators.cfc
```diff
@@ -41,23 +48,38 @@ component {
 	public array function existingEmailsUsingInvalidDomains( required string validDomains ) {
 		var delims    = ", #Chr( 10 )##Chr( 13 )#";
 		var domains   = ListToArray( Trim( arguments.validDomains ), delims );
-		var addresses = $getPresideObject( "email_template" ).selectData( distinct=true, selectFields=[ "from_address" ] );
+		var addresses = $getPresideObject( "email_template" ).selectData(
+			  distinct     = true
+			, selectFields = [ "id","from_address" ]
```
DominicWatson: We should filter here where `from_address` is not null. Also, we should select whether or not this is a system email (for check below).

system/services/validation/EmailCenterValidators.cfc
```diff
@@ -41,23 +48,38 @@ component {
...
+		);
 		var badaddresses = [];
 
 		for( var address in addresses ) {
 			if ( Len( address.from_address) ) {
-				var domain = _getDomainFromEmail( address.from_address );
-				if ( !ArrayFindNoCase( domains, domain ) ) {
-					ArrayAppend( badaddresses, address.from_address );
+				if ( _getSystemEmailTemplateService().templateExists( address.id ) ) {
```
DominicWatson: We should only run this check where the email is marked as a system email. Otherwise, we'll be unsetting the from address for all editorial emails with a from address I think?!

DominicWatson (changes requested):

PR #1203, opened 2023-01-03
PRESIDECMS-2437 typetoconfirmdelete not applicable on usermanager websiteusermanager

teonator:


DominicWatson:
Fab, this works as a "quick" "fix" for now. Longer term, we should just migrate everything to full datamanager.

PR #1204, opened 2023-01-05
PRESIDECMS-2438 relationship too complex error when quick edit rules conditions filter

karhong:


DominicWatson:
Nice one, merging for QA

PR #1205, opened 2023-01-09
PRESIDECMS-2440 ensure that having fields are unique

choontat-pixl8:


DominicWatson:
great stuff, thanks @choontat-pixl8 . Merging for QA now

PR #1206, opened 2023-01-13
PRESIDECMS-2443 create admin system alert for breaking rule engine rules

pixl8-brayden:


system/handlers/admin/systemAlerts/invalidRuleEngineRules.cfc
```diff
@@ -0,0 +1,51 @@
...
+	private void function runCheck( required systemAlertCheck check ) {
+		var invalidRules  = [];
+		var ruleFilter    = {};
+		var existingAlert = systemAlertsService.getAlert( type="invalidRuleEngineRules" );
+
+		if ( arrayLen( existingAlert.data.invalidRules ?: [] ) ) {
+			ruleFilter.id = existingAlert.data.invalidRules;
+		}
+
+		var allRules = getPresideObject( "rules_engine_condition" ).selectData( filter=ruleFilter, selectFields=[ "id", "context", "filter_object", "expressions" ] );
```
sebduggan: Not sure about the approach here. If an alert has been raised, then re-running the check will only ever recheck the previously found errors, and not find any new ones.

If we were to use the "references()" functionality to get all condition IDs:
* calling runCheck() with no reference would then get all references and run each in turn
* calling runCheck() with a reference would run the check for a specific condition, a more efficient way when adding/editing/deleting conditions
* we would have a separate alert for each failing condition

DominicWatson: Yes, the point is that the only time a condition/filter will be expected to be invalid is a new version of the software that means missing fields/expressions, etc.

So: on startup, we'd want to scan all
During running: check individual ones after they're edited (and possibly when inserted)

system/handlers/admin/systemAlerts/invalidRuleEngineRules.cfc
```diff
@@ -0,0 +1,59 @@
...
+			if ( !isRuleValid ) {
+				arrayAppend( invalidRules, rule.id );
+			}
+		}
+
+		if ( arrayLen( invalidRules ) ) {
+			check.fail();
+			check.setData( { invalidRules=invalidRules } );
+		} else {
+			check.pass();
```
sebduggan, line 42: A check object passes by default, so the "else" clause here is redundant.

system/handlers/admin/systemAlerts/invalidRuleEngineRules.cfc
```diff
@@ -0,0 +1,59 @@
...
+	property name="systemAlertsService" inject="SystemAlertsService";
+
+	private void function runCheck( required systemAlertCheck check ) {
+		var invalidRules  = [];
+		var ruleFilter    = "";
+		var ruleParams    = {};
+		var existingAlert = systemAlertsService.getAlert( type="invalidRuleEngineRules" );
+
+		if ( arrayLen( existingAlert.data.invalidRules ?: [] ) ) {
+			ruleFilter = "id in (:id) OR datecreated >= :datecreated";
```
sebduggan, line 13: The date filter is only applied if there is an existing alert. So, if there is no existing alert, all rules will be checked every time this runs, which is not ideal! But all rules should be checked on startup, as the available rules may have changed...

This requires a small extra piece in the main system alerts code which I'll add in now: 
- A check should know how it was triggered (e.g. startup, manual, etc)
- A check should know when it eas last run, whether or not an alert was raised

PR #1208, opened 2023-01-19
PRESIDECMS-2444 Formbuilder matrix throw error when row question has more than 100 chars

teonator:


DominicWatson:
I guess the question here is, given that the subreference for matrix is being based on the editorial _question_ - can we guarantee it will be 200 chars or less?

Am wondering if we can't use `Hash( question )` when saving it? (of course, this would require some backward compatibility checks for where we're searching against it). 

Or is the question length limited to 200 for matrix questions?

PR #1211, opened 2023-01-24
PRESIDECMS-2448 Refactor auto rules generation for related properties

DominicWatson:
Refactor how related property auto rules are implemented.
    
Instead of expecting all other rules expressions to understand this possibility - wrap any generated expressions of this kind in a wrapper logic that takes care of making an exists () subquery to isolate all the filter logic.
    
This means that child expressions need know nothing at all of chaining expressions, etc. and should keep queries far simpler.


DominicWatson:
Closing as moving to different branch

PR #1217, opened 2023-02-03
Feature preside 1809 incorrect maxlength default in cms properties

chunwengpixl8:
Currently it is
validation.maxLength.default=Must be no more than {1} characters

It should’ve been
validation.maxLength.default=Must be no more than {2} characters


DominicWatson:
Thanks @chunwengpixl8 - this is a little suspicious though. It must have worked at some point, and all the other related validators are also using `{1}` - I'll take a bit more of a look into what is causing that change.

DominicWatson:
Yes, looking at this more closely shows that is would expect to be `{1}`. Are you saying that it should be `{2}` because that _works_? or because you'd expect it be this due to the logic you are expecting?

I think we need to get to the bottom of why the `maxLength` value is not being passed as the first parameter to translateResource.

chunwengpixl8:
Hi @DominicWatson , I have looked at why is {1} not the number of characters.
This is because of the maxlength ruleset is having fieldLabel.
If you look at ValidationEngine$validate() line 81, you can see that it is actually prepending the name of the field to the params, that is why the number of chatacters become {2}.

DominicWatson:
Aha, next step - *git blame!* The issue here then is that *ArrayPrepend* is incorrect NOT that we should be using *{2}*. Caused by https://presidecms.atlassian.net/browse/PRESIDECMS-2369 that did some "tidying up" (https://github.com/pixl8/Preside-CMS/commit/cb9b4b9a72038bab41d09cb40e957beb4b812298#diff-3f27c16bccd571fc358901628693114a14900430c98bf34c7d76b059b6f3d35f)

Am closing this and will create a new fix branch addressing the route issue. Thanks for the extra investigation @chunwengpixl8

PR #1218, opened 2023-02-03
PRESIDECMS-2455 Datamanager clone record many to many field not prepopulated

teonator:


DominicWatson:
This feels like its addressing a symptom, rather than the cause and could cause other potentially unwanted side effects.

The root of the issue should be found in `handlers/admin/DataManager.cfc$_cloneRecordForm()`. i.e. is the field listed here as cloneable and if so, is its data populated? If not, this is where the fix needs to be.

teonator:
The field is listed as cloneable. The source ID field is ignored from cloneable list for many-to-many selector to populate data. Added ignoreIdField flag to include ID field for clone record form.

DominicWatson:
Nice one, let's get into QA

PR #1220, opened 2023-02-13
PRESIDECMS-2460: Option to preserve query strings on URL redirects

smoufa:


system/i18n/preside-objects/url_redirect_rule.properties
```diff
@@ -17,4 +17,7 @@ field.redirect_to_link.title=Redirect to
 field.redirect_to_link.help=Choose or add a link that the request will be redirected to when matching the incoming URL pattern
 
 field.redirect_type.title=Redirect type
-field.redirect_type.help=Whether the redirect is a permanent or temporary redirect. If in doubt, choose the default 'permanent' redirect.
\ No newline at end of file
+field.redirect_type.help=Whether the redirect is a permanent or temporary redirect. If in doubt, choose the default 'permanent' redirect.
+
+field.keep_query_string.title=Preserve URL parameters
+field.keep_query_string.help=If true, any URL parameters found at the incoming URL will be added to the target URL. Do not use this option if the target URL uses any parameters as a redirect destination, e.g. /target.html?dest=https://1.2.3.4/
```
sebduggan, line 23: Why not if the target has a querystring? The logic in the handler seems to manage this scenario...

smoufa: I think I read that there might be security problems if the target itself redirects users based on a passed parameter. Maybe I didn't think this through.

PR #1222, opened 2023-02-17
PRESIDECMS-2447: Optionally don't send scheduled exports to recipients if empty

smoufa:
* Option to omit sending scheduled exports when empty
* New field in scheduled export email with number of records in export


DominicWatson:
Nice work @smoufa . Merged for internal QA and hope to include in next release.

PR #1225, opened 2023-02-23
PRESIDECMS-2374: Add auto expression to check for outdated translations

smoufa:
Adds an auto expression to multilingual objects to allow filtering for translations that are older than the original content


DominicWatson:
awesome work @smoufa . Merging into the upcoming release.

PR #1226, opened 2023-03-01
Feature presidecms 2466 contact sent email preview width

john5onCheng:


sebduggan:
Rather than do this, we should ensure that the CSS and JS which goes with the log listing is used wherever that listing is called (it's not just a case of setting a width...)

PR #1235, opened 2023-03-17
Feature presidecms 2471 saved exports not showing the filter

Pixl8RusselCole:


system/handlers/admin/datamanager/saved_export.cfc
```diff
@@ -211,6 +221,8 @@ component {
 			  templateId = ( args.record.template ?: "" )
 			, objectName = rc.filterObject
 		);
+
+		args.additionalArgs.fields.filter.deleted=true;
```
DominicWatson: I don't *believe* this will have any effect at this point as `deleted` is only processed when merging form definitions, not when rendering them. Is it absolutely necessary?

DominicWatson (changes requested):

PR #1250, opened 2023-04-10
PRESIDECMS-2493 Prevent grid clickable link for formbuilder response

teonator:


system/assets/js/admin/presidecore/preside.clickable.tableRows.js
```diff
@@ -43,7 +43,7 @@
 			var $firstLink = $( this ).find( 'a.row-link:first' );
 
 			if ( !$firstLink.length ) {
-				$firstLink = $( this ).find( 'a:first' );
+				$firstLink = $( 'a', this ).not( '.formbuilder-response a' ).first();
```
DominicWatson: Rather than this hardcoded exception, is there a way we can assign `row-link` to the links in the response rows that we _do_ want clickable?

DominicWatson (changes requested):

system/assets/js/admin/presidecore/preside.clickable.tableRows.js
teonator: Update formbuilder submission action link with suggested `row-link` class.

PR #1265, opened 2023-05-08
PRESIDECMS-2512 Formbuilder multiple question w comma value

teonator:


system/handlers/formbuilder/item-types/CheckboxList.cfc
```diff
@@ -17,19 +17,24 @@ component {
...
-		var values    = ListToArray( args.itemConfiguration.values, chr(13) & chr(10) );
-		var labels    = ListToArray( args.itemConfiguration.labels, chr(13) & chr(10) );
+		var response = args.response ?: "";
 
-		var renderLabels = [];
-		for ( var response in responses ) {
-			var ix = values.find( response );
-			if ( ix>0 ) {
-				renderLabels.append( labels[ix] ?: values[ix] );
+		if ( isEmptyString( response ) ) {
```
DominicWatson: I think this should be `args.response`?

system/handlers/formbuilder/item-types/CheckboxList.cfc
```diff
@@ -17,19 +17,24 @@ component {
...
+			return "";
+		}
+
+		var values = ListToArray( args.itemConfiguration.values, chr(13) & chr(10) );
+		var labels = ListToArray( args.itemConfiguration.labels, chr(13) & chr(10) );
+
+		var data = [];
+
+		for ( var i=1; i<=ArrayLen( values ); i++ ) {
+			if ( Find( values[ i ], response ) ) {
```
DominicWatson: This could cause problems if one value is contained within another. We'll need something more robust here I think. If `response` (`args.response`?) is a comma list, then could do something like:

```cfc
var searchResponse = ",#args.response#,";
// ...
if ( Find( ",#values[ i ]#,", searchResponse ) {
}
```
**Feedback applies throughout the PR**

DominicWatson (changes requested):

system/handlers/formbuilder/item-types/CheckboxList.cfc
teonator: Updated.

system/handlers/formbuilder/item-types/CheckboxList.cfc
teonator: Updated.

system/handlers/formbuilder/item-types/CheckboxList.cfc
```diff
@@ -17,41 +17,45 @@ component {
...
 	}
 
 	private string function getQuestionDataType( event, rc, prc, args={} ) {
 		return "shorttext";
 	}
+
+	private array function _renderResponses( event, rc, prc, args={}, boolean useLabel=false ) {
+		var searchResponse = ",#args.response#,";
+
+		if ( isEmptyString( searchResponse ) ) {
```
DominicWatson: This will never be an empty string ;) `args.response` might be though.

system/handlers/formbuilder/item-types/CheckboxList.cfc
```diff
@@ -17,41 +17,45 @@ component {
...
+		var values = ListToArray( itemConfig.values ?: "", chr(13) & chr(10) );
+		var labels = ListToArray( itemConfig.labels ?: "", chr(13) & chr(10) );
+
+		for ( var i=1; i<=ArrayLen( values ); i++ ) {
+			if ( Find( ",#values[ i ]#,", searchResponse ) ) {
+				var renderResponse = useLabel ? ( labels[ i ] ?: values[ i ] ) : values[ i ];
+
+				ArrayAppend( matchResponses, renderResponse );
+
+				searchResponse = REReplace( Replace( searchResponse, ",#values[ i ]#,", "" ), "^,?(.*?),?$", ",\1," );
```
DominicWatson: I think this can just be:

```cfc
searchResponse = Replace( searchResponse, ",#values[ i ]#,", "," );
```

i.e. replace it with a single comma, rather than empty string.

DominicWatson (changes requested):

PR #1287, opened 2023-05-29
PRESIDECMS-2535 improve admin user group role picker

pixl8-brayden:


system/assets/css/admin/specific/rolepicker/rolepicker.less
```diff
@@ -0,0 +1,22 @@
+@import url( 'globals.less' );
```
DominicWatson, line 1: Given this is specific css for a role picker, it would make sense to wrap the whole thing in a containing div with a `preside-role-picker` class so that the css remains specific? So add that class to the css + add a wrapping div to the markup.

system/services/security/PermissionService.cfc
```diff
@@ -52,6 +52,33 @@ component displayName="Admin permissions service" {
...
+
+		if ( !IsNull( local.fromCache ) ) {
+			return fromCache;
+		}
+
+		var grouped = {};
+		var roles   = listRoles();
+
+		for ( var role in roles ) {
+			var roleGroup = $translateResource( uri="roles:#role#.group", defaultValue="" );
```
DominicWatson: We can't cache translations unless we include the current language in the cache key :). Otherwise multilingual admin fails.

Current language can be got from i18n service I believe.

DominicWatson (changes requested):
Looking good, see feedback comments in code.

PR #1302, opened 2023-06-13
PRESIDECMS-2556 bring better datamanager record view into core

sebduggan:


DominicWatson (changes requested):
Looking good, but looks like some tmp file got committed accidentally

image

PR #1309, opened 2023-06-19
PRESIDECMS-2494 Formcontrol HTML attributes

teonator:


system/helpers/htmlHelpers.cfm
```diff
@@ -18,3 +18,24 @@
 	<cfargument name="stringValue" type="string" required="true" /><cfsilent>
 	<cfreturn IsTrue( ReFind(  "<[^>]*>",stringValue ) ) />
 </cfsilent></cffunction>
+
+<cffunction name="renderForHTMLAttributes" access="public" returntype="string" output="false">
```
DominicWatson: `renderHtmlAttributes` is a better name but I can see it clashing with an existing function we have privately in Pixl8... 

The form of that helper is:

```cfc
public string function renderHtmlAttributes( required struct attribs ){
	var rendered = [];

	for ( var key in arguments.attribs ) {
		rendered.append( '#key#="#arguments.attribs[ key ]#"' );
	}

	return rendered.toList( " " );
}
```

I wonder if we could refactor the helper here to be compatible with that and update that helper internally with these changes? Something like:

```cfc
public string function renderHtmlAttributes( struct attribs={}, string attribNames="", string attribValues="", string attribPrefix="" ){
	var rendered = [];

	// convert separated names+values to struct
	var names  = ListToArray( arguments.attribNames );
	var values = ListToArray( arguments.attribValues );
	for ( var i=1; i<=ArrayLen( names ); i++ ) {
		arguments.attribs[ names[ i ] ] = values[ i ] ?: "";
	}

	for ( var key in arguments.attribs ) {
		var encodedName  = EncodeForHTMLAttribute( arguments.attribPrefix & key );
		var encodedValue = EncodeForHTMLAttribute( arguments.attribs[ key ] );
		
		ArrayAppend( rendered, '#encodedName#="#encodedValue#"' );
	}

	return ArrayToList( rendered, " " );
}
```

system/helpers/htmlHelpers.cfm
```diff
@@ -18,3 +18,24 @@
...
+	<cfargument name="htmlAttributeNames"  type="string" default="" />
+	<cfargument name="htmlAttributeValues" type="string" default="" />
+	<cfargument name="htmlAttributes"      type="struct" default="#StructNew()#" />
+	<cfargument name="htmlAttributePrefix" type="string" default="data-" /><cfsilent>
```
DominicWatson: As per example above, I think an empty string is a better default here. Allows for more flexibility by default.

DominicWatson (changes requested):
Nice concept. I've added some feedback on the helper method. I think we can rename and refactor a little (would also require updated references).

Also, I think that using `extraAttributes = renderHtmlAttributes( ... )` is a sensible nomenclature throughout all the individual control files. Makes it clear that these are in addition to anything standard.

PR #1320, opened 2023-07-14
PRESIDECMS-2656 dependabot alerts tough cookie prototype pollution vulnerability

pixl8-brayden:
run rpm update


DominicWatson:
Nice, looks like removing even more redundant dependencies too.

PR #1327, opened 2023-07-24
PRESIDECMS-2665 better indexes on email logs

DominicWatson:
To help performance when a system sends a lot of email - we need to optimize indexes, and other aspects. Here, we are improving the indexes on the send log and activity tables and refactoring the stats screen calculation to reduce repeated queries (i.e. do all the time series queries in a single query)


system/services/email/EmailTemplateService.cfc
```diff
@@ -843,6 +845,7 @@ component {
 		  required string templateId
 		,          string dateFrom = ""
 		,          string dateTo   = ""
+		,          numeric timePoints = 1
```
sebduggan: What's this doing? Doesn;t seem to be used...

PR #1328, opened 2023-07-24
PRESIDECMS-2592: Add RecordMatchesId auto-expression

smoufa:
...to filter for specific records


DominicWatson:
Extra thanks for the passing tests!

PR #1329, opened 2023-07-25
PRESIDECMS-2667 align attribute missing on image picker renderer causing image not align correctly on outlook

s-ven-315:


DominicWatson:
Thanks @s-ven-315  - we'll need to be a little careful with testing in browsers just to double check that adding this attribute everywhere just for outlook isn't going to cause a problem. Have made QA note of this.

PR #1332, opened 2023-07-31
PRESIDECMS-2676 use inspectTemplates if available

zspitzer:
https://presidecms.atlassian.net/browse/PRESIDECMS-2676


zspitzer:
hang on that doesn't quite work as intended

DominicWatson:
Nice one, added a comment re `StructFindKey()` vs `StructKeyExists()` - but a nice addition which we can merge in next release.

zspitzer:
@DominicWatson yeah, I have cleaned that up

PR #1337, opened 2023-08-08
PRESIDECMS-2684 Formbuilder formitem without question

teonator:


system/handlers/admin/datamanager/formbuilder_question.cfc
```diff
@@ -244,6 +244,21 @@ component extends="preside.system.base.AdminHandler" {
...
+		var logger  = args.logger  ?: NullValue();
+		var index   = 1;
+
+		for ( var record in records ) {
+			if ( formBuilderQuestionService.questionIsInUse( questionId=record.id ) ) {
+				logMessage( logger, "warn", translateResource( uri="preside-objects.formbuilder_question:batchAction.question.in.use.warning", data=[ record.label ] ) );
+
+				QueryDeleteRow( args.records, index );
+			}
+			index++;
```
sebduggan: This loop is problematic. If you delete row 1, all the rows shift up, so if you then delete row 2, you've actually deleted the original row 3.

For illustration, run this in [trycf.com](trycf.com):

```
<cfscript>
 	testQuery = queryNew( "number" , " numeric" , { number: [ 1,2,3,4,5,6,7,8,9,10 ] } );
 	dump(testQuery); 
 	
 	QueryDeleteRow( testQuery, 1 );
 	QueryDeleteRow( testQuery, 2 );
 	
 	dump(testQuery); 
</cfscript>
```

Simplest solution would be to only increase the index if you haven't deleted the row (put it in the else{} ).

PR #1343, opened 2023-08-16
Feature presidecms 2687 preside validation field param translation

mookeshNanda:
Added paramsFieldLabel struct as part of rule which later be used as the label field for the param.
Note, currently the paramsFieldLabel will have value only when there is "field" param being used for the rule.

### DIFF
--- a/system/services/validation/PresideFieldRuleGenerator.cfc
+++ b/system/services/validation/PresideFieldRuleGenerator.cfc
@@ -86,8 +86,17 @@ component output="false" singleton=true {
 
 					if ( StructKeyExists( field, "rules" ) ) {
 						for( rule in field.rules ){
-							rule.fieldName  = field.name;
-							rule.fieldLabel = field.label;
+							rule.fieldName        = field.name;
+							rule.fieldLabel       = field.label;
+							rule.paramsFieldLabel = {};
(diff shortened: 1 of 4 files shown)


system/services/validation/PresideFieldRuleGenerator.cfc
```diff
@@ -86,8 +86,17 @@ component output="false" singleton=true {
 
 					if ( StructKeyExists( field, "rules" ) ) {
 						for( rule in field.rules ){
-							rule.fieldName  = field.name;
-							rule.fieldLabel = field.label;
+							rule.fieldName        = field.name;
+							rule.fieldLabel       = field.label;
+							rule.paramsFieldLabel = {};
+							if ( !StructIsEmpty( rule.params ?: {} ) && StructKeyExists( rule.params, "field" ) ) {
+								for ( var f in fieldset.fields ) {
```
sebduggan: Will this only work if the referenced field is in the same fieldset?

mookeshNanda: Hi @sebduggan ,

Yup, it would only work if the field is in the same fieldsets as I have not seen reference field being defined on different fieldset .
Probably the referenced field could potentially defined on different fieldet.
I could update it to look through all the fieldsets, but wondering if we should also look through all the tabs, tho I doubt reference field going to be defined as such.

Let me know your thought.

PR #1349, opened 2023-08-22
PRESIDECMS-1845 getLoggedInUserDetails not updated when website user data is updated

teonator:


system/services/websiteUsers/WebsiteLoginService.cfc
```diff
@@ -569,6 +569,17 @@ component displayName="Website login service" {
 		return false;
 	}
 
+	public void function setUserSessionById( required string id ) {
```
DominicWatson: This feels somewhat dangerous. If someone is able to edit another user, then will they not then become logged in as that user with this change?

I think it may be preferable to implement specific session refresh in any implementations that perform editing of self through the frontend rather than globally when updating any user record.

DominicWatson (changes requested):

system/services/websiteUsers/WebsiteLoginService.cfc
teonator: Replace the interceptor with helper.

PR #1351, opened 2023-08-24
PRESIDECMS-1983 CSRF token validation auth forms

teonator:


system/handlers/Login.cfc
```diff
@@ -28,8 +30,6 @@ component {
 			setNextEvent( url=postLoginUrl );
 		}
 
-		announceInterception( "onLoginFailure"  );
```
sebduggan: Should this have been removed?

PR #1361, opened 2023-09-15
PRESIDECMS-2708 better cache headers for assets

DominicWatson:
Make use of `private` cache-control with shorter timeouts for restricted assets.

### DIFF
--- a/system/handlers/core/AssetDownload.cfc
+++ b/system/handlers/core/AssetDownload.cfc
@@ -8,10 +8,14 @@ component {
 	property name="publicCacheAge"               inject="coldbox:setting:assetManager.cacheExpiry.public";
 	property name="privateCacheAge"              inject="coldbox:setting:assetManager.cacheExpiry.private";
 
-	public function asset( event, rc, prc ) output=false {
+	public function asset( event, rc, prc ) {
 		announceInterception( "preDownloadAsset" );
 
-		var isRestricted      = _checkDownloadPermissions( argumentCollection=arguments );
+		var permissionSettings = _getPermissionSettings( argumentCollection=arguments );
+		if ( permissionSettings.restricted ) {
+			_checkDownloadPermissions( argumentCollection=arguments, permissionSettings=permissionSettings );
+		}
+
 		var assetId           = rc.assetId      ?: "";
 		var versionId         = rc.versionId    ?: "";
 		var derivativeName    = rc.derivativeId ?: "";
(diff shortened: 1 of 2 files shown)


system/handlers/core/AssetDownload.cfc
```diff
@@ -178,54 +184,55 @@ component {
 		return ReReplace( arguments.assetTitle, "\.#arguments.extension#$", "" ) & "." & arguments.extension;
 	}
 
-	private void function _checkDownloadPermissions( event, rc, prc ) output=false {
+	private boolean function _checkDownloadPermissions( event, rc, prc ) output=false {
```
sebduggan: Is this the best name for the method if it's now returning a boolean? Doesn't really explain what the boolean signifies...

system/handlers/core/AssetDownload.cfc
```diff
@@ -178,54 +184,55 @@ component {
...
-				var hasPerm = event.isAdminUser() && hasCmsPermission(
-					  permissionKey       = "assetmanager.assets.download"
-					, context             = "assetmanagerfolder"
-					, contextKeys         = permissionSettings.contextTree
-					, forceGrantByDefault = IsBoolean( permissionSettings.grantAcessToAllLoggedInUsers ) && permissionSettings.grantAcessToAllLoggedInUsers
-				);
-				if ( hasPerm ) { return; }
+				return true;
+			}
+			var hasPerm = event.isAdminUser() && hasCmsPermission(
```
sebduggan: Not sure what this check is trying to achieve, but it will never be run - we've already established that the user is not an admin user...

system/handlers/core/AssetDownload.cfc
```diff
@@ -178,54 +184,55 @@ component {
...
+				, context             = "asset"
+				, contextKeys         = permissionSettings.contextTree
+				, forceGrantByDefault = IsBoolean( permissionSettings.grantAcessToAllLoggedInUsers ) && permissionSettings.grantAcessToAllLoggedInUsers
+			)
+			if ( !hasPerm ) {
+				event.accessDenied( reason="INSUFFICIENT_PRIVILEGES" );
 			}
 		} else {
-
 			hasPerm = hasCmsPermission(
```
sebduggan, line 207: The hasPerm variable has not been var scoped in this branch of the if/else

sebduggan (changes requested):

system/handlers/core/AssetDownload.cfc
DominicWatson: Yeah, is serving two purposes which is a little funky. i.e. it _does_ perform a check in that it will block the user. But having it also return a value. Bit weak as a pattern.

system/handlers/core/AssetDownload.cfc
DominicWatson: Yup, that was hungover from previous refactors I guess. Refactored now.

system/handlers/core/AssetDownload.cfc
DominicWatson: Have split things out into separate methods now.

PR #1362, opened 2023-09-18
PRESIDECMS-2709 preside utils rest api endpoint

pixl8-brayden:


system/config/Config.cfc
```diff
@@ -980,6 +980,14 @@ component {
 		};
 
 		settings.rest.authProviders.token = { feature = "restTokenAuth" };
+
+		settings.rest.apis[ "/presideUtils" ]   = {
+			  authProvider        = "presideApi"
+			, description         = "Utility REST API for external systems to interact with Preside"
+			, dataApiQueueEnabled = false
```
DominicWatson, line 987: Data API is not part of core Preside and cannot be used here. I think you just need to remove `, dataApiQueueEnabled = false`

system/handlers/rest/auth/PresideApi.cfc
```diff
@@ -0,0 +1,26 @@
...
+	private string function authenticate() {
+		var headers    = getHTTPRequestData( false ).headers;
+		var authHeader = headers.Authorization ?: "";
+		var token      = "";
+
+		try {
+			authHeader = toString( toBinary( listRest( authHeader, ' ' ) ) );
+			token      = ListFirst( authHeader, ":" );
+
+			if ( !token.trim().len() ) {
```
DominicWatson, line 13: `Len( Trim( token ) )`

system/handlers/rest/auth/PresideApi.cfc
```diff
@@ -0,0 +1,26 @@
...
+				throw( type="missing.token" );
```
DominicWatson, line 14: I think rather than throwing here, we should be setting `restResponse.setStatusText( "Missing authentication token" );` See https://github.com/pixl8/preside-ext-data-api/blob/stable/handlers/rest/auth/DataApi.cfc, for example.

Basically, if we confirm that token is valid, return the token. Otherwise, return empty string and set the reason for failure in the restResponse.setStatusText()

system/config/Config.cfc
DominicWatson: I think we also ought to have a feature flag on this that would be disabled by default. Either that, or we move this whole feature out into a different extension (e.g. our test data ext).

system/config/Config.cfc
```diff
@@ -980,6 +980,14 @@ component {
 		};
 
 		settings.rest.authProviders.token = { feature = "restTokenAuth" };
+
+		settings.rest.apis[ "/presideUtils" ]   = {
```
DominicWatson, line 984: Not sure about the naming here. Perhaps `/system/v1` (should include a version number so can make new versions in the future)?

DominicWatson (changes requested):

PR #1364, opened 2023-09-20
PRESIDECMS-2713 use faster Lucee 6 dbinfo when available

zspitzer:
https://presidecms.atlassian.net/browse/PRESIDECMS-2713

### DIFF
--- a/system/services/presideObjects/SqlSchemaSynchronizer.cfc
+++ b/system/services/presideObjects/SqlSchemaSynchronizer.cfc
@@ -51,6 +51,7 @@ component {
 				dbVersion.append( obj.sql.table.version );
 			}
 		}
+		StructDelete( variables, "columnCache" ); // cache no longer required, release memory
 		dbVersion.sort( "text" );
 		dbVersion = Hash( dbVersion.toList() );
 		if ( ( versions.db.db ?: "" ) neq dbVersion ) {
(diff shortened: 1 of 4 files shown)


tests/server.json
```diff
@@ -8,6 +8,9 @@
             "port":"9999"
         }
     },
+    "app":{
```
DominicWatson: While we should probably be using a matrix of versions for support, we shouldn't be making a 6 snapshot our main Preside test platform.

We should ensure this continues to work on 5 as a priority over running the main test suite in 6.

DominicWatson (changes requested):
Awesome looks great. Just one comment re changing the test suite Lucee version (we shouldn't be).

tests/server.json
zspitzer: sure, no problem, just wanted to see if it works with 6, was literally going back to mark this as a draft PR!

system/services/presideObjects/SqlSchemaSynchronizer.cfc
```diff
@@ -188,7 +188,7 @@ component {
 
 		for( colName in ListToArray( arguments.dbFieldList ) ) {
 			colMeta = arguments.properties[ colName ];
-			if( listFindNoCase( tableColumns, colName ) && _skipSync( colMeta ) ) {
+			if( _skipSync( colMeta ) && ArrayContainsNoCase( tableColumns, colName ) ) {
```
zspitzer, line 192: did this boolean skip check first as it's cheaper?

system/services/presideObjects/SqlSchemaSynchronizer.cfc
```diff
@@ -702,7 +702,29 @@ component {
 		return variables._tableCache[ arguments.dsn ][ arguments.tableName ] ?: false;
 	}
 
-	private query function _getTableColumns() {
+	private array function _getTableColumns() {
+		try {
+			if ( _getDbInfoService().hasModernDbInfo() ){
+				// with the faster version, we can fetch the entire schema in one call.
+				if ( !StructKeyExists( variables, "columnCache" )){
+					variables.columnCache =  _getTableColumnCache( argumentCollection = arguments );
```
zspitzer, line 711: it's my understanding this only runs once at startup? but it probably makes sense to purge this cache after the sync finishes, no need for a query with all the columns in the db to be left hanging in memory

zspitzer:
Been testing, this reduces the db sync time by about roughly 50% on my fast laptop, 12s to 7s

DominicWatson:
Great, thanks for the tidy up also. Will get this merged and in use internally for QA against existing Lucee.

PR #1367, opened 2023-09-21
PRESIDECMS-2559 Formbuilder form fields export import

teonator:


system/handlers/admin/FormBuilder.cfc
```diff
@@ -849,6 +848,87 @@ component extends="preside.system.base.AdminHandler" {
...
+
+		if ( canProgress ) {
+			arguments.progress.setProgress( 100 );
+		}
+
+		logMessage( logger, "info", "Finished import." );
+	}
+
+	public string function exportFormFieldsAction( event, rc, prc, args ) {
+		var fileName = formBuilderService.exportFormFields( formId=rc.id ?: "" );
```
DominicWatson: Filename should be the full filepath. This calling logic should not need to know that it is using the tmp directory.

system/handlers/admin/FormBuilder.cfc
```diff
@@ -849,6 +848,87 @@ component extends="preside.system.base.AdminHandler" {
...
+
+		if ( isEmptyString( fileName ) ) {
+			event.notFound();
+		}
+
+		header name="Content-Disposition" value="attachment; filename=""#fileName#""";
+		content reset=true file=getTempDirectory() & "/" & fileName deletefile=true type="application/xml";
```
DominicWatson: application/json?

DominicWatson (changes requested):

PR #1376, opened 2023-10-12
Feature presidecms 2724 email center link stat table slow query

mookesh:
https://presidecms.atlassian.net/browse/PRESIDECMS-2724

### DIFF
--- a/system/handlers/dbmigrations/emailTemplateActivityClickStatHash.cfc
+++ b/system/handlers/dbmigrations/emailTemplateActivityClickStatHash.cfc
@@ -1,0 +1,122 @@
+component {
+
+	private void function run() {
+		var presideObjectService = getController().getWirebox().getInstance( "presideObjectService" );
+		var dbInfoService        = getController().getWirebox().getInstance( "dbInfoService" );
+		var sqlRunner            = getController().getWirebox().getInstance( "sqlRunner" );
+		var objectName           = "email_template_send_log_activity";
+		var dbAdapter            = presideObjectService.getDbAdapterForObject( objectName );
+		var dsn                  = presideObjectService.getObjectAttribute( objectName, "dsn" );
+		var tableName            = presideObjectService.getObjectAttribute( objectName, "tablename" );
+		var dbInfo               = dbInfoService.getDatabaseVersion( dsn );
+		var escapedTableName     = dbAdapter.escapeEntity( tableName );
+		var updateHashCols       = [];
+		var addTempColSqlArray   = [];
+		var dropTempColSqlArray  = [];
+		var addTempIndexSql      = "";
+		var dropTempIndexSql     = "";
+
+		for ( var rawCol in [ "link", "link_title", "link_body" ] ) {
+			var escapedRawCol      = dbAdapter.escapeEntity( rawCol );
+			var escapedHashCol     = dbAdapter.escapeEntity( "#rawCol#_hash" );
+			var escapedHashTempCol = dbAdapter.escapeEntity( "#rawCol#_hash_temp" );
+
+			var addTempColSql  = "";
+			var dropTempColSql = "";
+
+			switch ( dbInfo.database_productName ) {
+				case "MySQL":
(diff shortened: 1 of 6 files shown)


system/handlers/dbmigrations/emailTemplateActivityClickStatHash.cfc
```diff
@@ -0,0 +1,122 @@
...
+
+		for ( var rawCol in [ "link", "link_title", "link_body" ] ) {
+			var escapedRawCol      = dbAdapter.escapeEntity( rawCol );
+			var escapedHashCol     = dbAdapter.escapeEntity( "#rawCol#_hash" );
+			var escapedHashTempCol = dbAdapter.escapeEntity( "#rawCol#_hash_temp" );
+
+			var addTempColSql  = "";
+			var dropTempColSql = "";
+
+			switch ( dbInfo.database_productName ) {
```
DominicWatson: There is an issue here in that this requires the database user to have DDL privileges. For production instances, it is advised that the db user for the application only has data privileges and is not able to update the schema (and this is the case at Pixl8, for example, which means this migration will fail).

Did you consider running something like the following instead?:

```sql
update #tablename#
set #hashCol# = md5(#sourceCol#)
where #hashCol# is null and #sourceCol# is not null
```

DominicWatson (changes requested):
Love this + especially the research into it. I think the db migration needs refactoring though.

system/handlers/dbmigrations/emailTemplateActivityClickStatHash.cfc
mookesh: Yup, that make sense Dom. 
Have updated the dbmigration script. 

Also updated filter in the getLinkClickStats which was missing the null check.

system/services/email/EmailTemplateService.cfc
```diff
@@ -1202,15 +1208,47 @@ component {
...
-			  filter       = { "message.email_template"=arguments.templateId }
-			, selectFields = [ "count( 1 ) as click_count", "link", "link_title", "link_body" ]
-			, extraFilters = extraFilters
-			, autoGroupBy  = true
-			, orderBy      = "click_count desc"
+		var subQuerySelectFields      = [ "count(1) as click_count", "id" ];
+		var rawClickStatsSelectfields = [ "sendLog.click_count", "link", "link_title", "link_body" ];
+		var subQueryGroupBy           = "link_hash,link_title_hash,link_body_hash";
+		var rawClickStatsGroupBy      = "link_hash,link_title_hash,link_body_hash";
+		var dbInfo                    = _getDbInfoService().getDatabaseVersion( "preside" );
```
DominicWatson: While we do expect that most, if not all, preside apps have a DSN named "preside", it is not a requirement and we do not hardcode this anywhere else in core.

The general approach here is to use our db adapter pattern so:

```cfc
var adapter = $getPresideObject( "email_template_send_log_activity" ).getDbAdapter();
```

We then can/should use that adapter to help generate the SQL we need. In this case, it isn't going to have the methods and so we might need to add some. e.g.

```cfc
if ( !adapter.supportsGroupByWithAnyFields() ) { // rather than checking whether or not SQL server
   // prepare alternative sql
}
```

We can then add that method to the base adapter (returning true) and then have it return false in the SQL server adapters.

DominicWatson (changes requested):

system/services/email/EmailTemplateService.cfc
mookesh: Have added helper "supportsGroupByAnyField" on baseAdaptor.
Default is false as realised that Postgres does not support the groupBy with any fields either. MySql would return only if the sql_mode does not have "ONLY_FULL_GROUP_BY" .

Additionally, added another helper "supportsGroupByTextField" as MSSQL does not allow group by text field

DominicWatson:
Appreciate all the work here. However, closing due to an alternative solution that takes a wholly different approach.

PR #1384, opened 2023-10-25
PRESIDECMS-2721 data export ability to choose related object fields

pixl8-brayden:


system/handlers/admin/DataManager.cfc
```diff
@@ -3382,6 +3383,20 @@ component extends="preside.system.base.AdminHandler" {
 			, templateConfig     = dataExportTemplateService.getSubmittedConfig( exportTemplate, objectName )
 		};
 
+		var appEnableExpandExportFields = isTrue( dataExportDefaultConfig.expandManytoOneFields ?: "" );
+		var objEnableExpandExportFields = isTrue( presideObjectService.getObjectAttribute( objectName=args.objectName, attributeName="dataExportExpandManytoOneFields" ) );
+
+		args.expandNestedFields = appEnableExpandExportFields || objEnableExpandExportFields;
+		if ( !args.expandNestedFields ) {
+			for ( var field in args.selectFields ) {
+				writeDump(field);
```
sebduggan: Debug code!

system/handlers/admin/DataManager.cfc
```diff
@@ -3382,6 +3383,20 @@ component extends="preside.system.base.AdminHandler" {
...
+				if ( ListLen( field, "." ) > 1 ) {
+					args.expandNestedFields = true;
+					break;
+				}
+			}
+		}
+
```
sebduggan: Not sure we need any of the above logic. In here, if we're using _exportDataAction, shouldn't expandNestedFields always be true?

The custom exports send their select fields direct to the service, which is what causes the issue as they might already be aliased...

system/services/dataExport/DataExportService.cfc
```diff
@@ -699,6 +715,31 @@ component {
...
+		var processed = {
+			  shouldExpand = false
+			, expandFields = []
+		};
+
+		if ( StructKeyExists( arguments.propAttributes, "dataExportExpandFields" ) ) {
+
+			if ( IsBoolean( arguments.propAttributes.dataExportExpandFields ) ) {
+				processed.shouldExpand = arguments.propAttributes.dataExportExpandFields;
+			} else if ( ListLen( arguments.propAttributes.dataExportExpandFields ) > 0 ) {
```
sebduggan: I think this simplifies to:

`} else if ( Len( arguments.propAttributes.dataExportExpandFields ) ) {`

PR #1388, opened 2023-11-08
PRESIDECMS-2735 Objects in datamanager listing are shown even though they dont have any permission

teonator:


system/handlers/admin/DataManager.cfc
```diff
@@ -39,6 +39,16 @@ component extends="preside.system.base.AdminHandler" {
 
 		prc.objectGroups = dataManagerService.getGroupedObjects();
 
+		for ( var g=1; g<=ArrayLen(  prc.objectGroups ); g++ ) {
```
DominicWatson: I just checked to see if this could be refactored into the `getGroupedObjects()` call and it appears that we are already checking for perms there: https://github.com/pixl8/Preside-CMS/blob/498f1ca9981af1ec09d24b2ce76b11916cd0853d/system/services/admin/DataManagerService.cfc#L72

Do we _definitely_ need to do this here? Or might just adding the "navigation" key as below fix the issue?

teonator: Removed handler loop and add navigate permission check in service.

PR #1396, opened 2023-11-22
PRESIDECMS-2375: Alternative site tree page listing

smoufa:


system/views/admin/sitetree/index.cfm
```diff
@@ -1,60 +1,39 @@
...
-	prc.pageIcon  = "sitemap";
-	prc.pageTitle = site.name ?: translateResource( "cms:sitetree" );
+	prc.pageIcon    = "sitemap";
+	prc.pageTitle   = site.name ?: translateResource( "cms:sitetree" );
+	rc.id           = "page";
+	args.objectName = "page";
 
 	activeTree          = prc.activeTree ?: [];
 	trashCount          = prc.trashCount ?: 0;
+	activeTab           = rc.tab ?: "sitetree";
```
sebduggan, line 12: Might be good to have a bit of validation on the provided tab; if it's not one of the expected two, then default to "sitetree"

system/preside-objects/core/page.cfc
```diff
@@ -1,5 +1,10 @@
 /**
  * The page object represents the core data that is stored for all pages in the site tree, regardless of page type.
+ *
+ * @datamanagerEnabled              true
+ * @datamanagerGridFields           page_type,title,parent_page,active,datecreated,datemodified
+ * @datamanagerDefaultSortOrder     sort_order,parent_page,active,title,datecreated,datemodified
```
sebduggan: This sort order doesn't seem to have any useful purpose. As specified, all pages that are 1 in their own hierarchy will be followed by all that are 2  in theirs, and so on. This gives a really jumbled and meaningless default order.

I'd suggest having a single sort order field, _hierarchy_sort_order. This will basically give the same order as in the tree view, but expanded and flattened, and so giving a logical connection between the two views. From there, the user is free to sort on date, title, etc.

Coupled with this, it might be informative to add _hierarchy_depth into the displayed columns:

image

system/views/admin/sitetree/index.cfm
smoufa: Makes sense. Done.

system/preside-objects/core/page.cfc
smoufa: Very good points. The order is much more aligned to the tree ordering that way. I changed that and also added the tree depth column.

smoufa:
The page object's read operation is now disallowed so that only the edit and history actions are offerered in the grid view.

PR #1400, opened 2023-12-19
PRESIDECMS-2754 allow clear page cache via custom url and page types

karhong:


system/services/siteTree/SiteTreeService.cfc
```diff
@@ -1357,17 +1357,32 @@ component {
 		return page.id ?: "";
 	}
 
-	public void function clearPageCache( required string pageId ) {
-		var pageUrl    = ReReplace( $getRequestContext().buildLink( page=arguments.pageId ), "^https?://.*?/", "/" );
+	public void function clearAllCaches() {
+		_getCachebox().clearAll();
+		announceInterception( "onClearCaches", {} );
```
DominicWatson: This needs `$` in front of `$announceInterception()`?

DominicWatson (changes requested):

system/services/siteTree/SiteTreeService.cfc
karhong: Arh yes, nice catch

PR #1404, opened 2024-01-09
Feature presidecms 2565 fk field with same name as object leads to errors filtering

Pixl8RusselCole:


system/services/presideObjects/PresideObjectService.cfc
```diff
@@ -2889,6 +2889,14 @@ component displayName="Preside Object Service" {
 
 				if ( objectExists( objName ) ) {
 					cols = _getObject( objName ).meta.properties;
+
+					if ( !structKeyExists( cols, ListLast( key, ". ") ) ) {
+						cols = arguments.columnDefinitions;
+					} else {
+						if ( objName!=arguments.objectName && structKeyExists( arguments.columnDefinitions, ListLast( key, "." ) ) ) {
```
DominicWatson, line 2896: This doesn't feel right. We should be able to work with both and I think the answer is that the query param names would need to be made unique is all?

DominicWatson (changes requested):

PR #1407, opened 2024-01-15
PRESIDECMS-2632 task to purge expired forms submission

pixl8-brayden:


system/services/formbuilder/FormBuilderService.cfc
```diff
@@ -1357,6 +1357,46 @@ component {
 		}
 	}
 
+	public boolean function deleteExpiredSubmissions( any logger=NullValue() ) {
```
Pixl8RusselCole, line 1360: I don't think you need =NullValue() here
logMessage will handle it anyway if it is null

system/i18n/preside-objects/formbuilder_form.properties
```diff
@@ -31,4 +31,8 @@ field.login_required_content.help=Optional content that will be displayed in pla
 field.access_denied_content.title=Access denied content
 field.access_denied_content.help=Optional content that will be displayed in place of the form if the user does not meet the access condition
 field.notification_enabled.title=Enable system notifications
-field.notification_enabled.help=If checked, form submissions will trigger an admin notification. You can subscribe to these notifications in your admin profile settings.
\ No newline at end of file
+field.notification_enabled.help=If checked, form submissions will trigger an admin notification. You can subscribe to these notifications in your admin profile settings.
+
+fieldset.submission_retention.title=Submission retention
+field.submission_remove_after.title=Remove after [x] days
+field.submission_remove_after.help=How long, in days, to remove the form submission. If left blank or 0, the system will never remove any submissions.
```
Pixl8RusselCole: Maybe?
After how long, in days, to remove the form submission. If left blank or 0, the system will never remove any submissions. 

Possibly add this?
Important: data from removed form submissions cannot be recovered

PR #1410, opened 2024-01-18
PRESIDECMS-2468 antisamy converting quotes

sebduggan:


system/config/Config.cfc
```diff
@@ -973,7 +973,7 @@ component {
 		settings.antiSamy = {
 			  enabled                 = true
 			, policy                  = "preside"
-			, bypassForAdministrators = true
+			, bypassForAdministrators = false
```
DominicWatson: I think we can't change this default.

DominicWatson (changes requested):
Looking good - just comment on the admin default.

PR #1417, opened 2024-02-08
PRESIDECMS-2780: Interception points for preRunCustomization and postRunCustomization

jjannek:


jjannek:
checking test suite now

PR #1422, opened 2024-02-19
PRESIDECMS-2790 handle utf8 content in linkPicker resources

zspitzer:
I'm mot 100% confident of the approach, but it addresses the described problem

https://presidecms.atlassian.net/browse/PRESIDECMS-2790


DominicWatson:
Looks good to me :+1:

zspitzer:
Added a v: 2 flag to the data object, as attempting to decode an un-encoded string can throw errors

PR #1423, opened 2024-02-19
PRESIDECMS-2791 missing prepost sortrecordsaction data manager customisation

pixl8-brayden:


system/handlers/admin/DataManager.cfc
```diff
@@ -1382,11 +1382,28 @@ component extends="preside.system.base.AdminHandler" {
 
 		_checkPermission( argumentCollection=arguments, key="edit", object=objectName );
 
+		var args = arguments;
```
DominicWatson: What is the purpose of passing the arguments to the customisation (customisation handlers will always receive event, rc, + prc so just doubling them up in the args here)? I'd suggest it would make sense this way:

```cfc
var args = {
      objectName = prc.objectName ?: ""
   ,  sortedIds  = ListToArray( rc.ordered ?: "" )
};
```

Can then use `args.sortedIds` when saving the records.

DominicWatson (changes requested):

PR #1424, opened 2024-02-19
PRESIDECMS-2792 frontend admin toolbar modes

sebduggan:


system/forms/system-config/admin-users.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
DominicWatson: I'm not sure "Admin users" is the right category here (given we also have admin login security). Could we make this a "Frontend editing" category instead?

system/views/general/_adminToolbar.cfm
```diff
@@ -117,6 +126,38 @@
 			</div>
 		</div>
 
+		<script>
+			( function(){
```
DominicWatson, line 130: old skool. love it.

DominicWatson (changes requested):

PR #1437, opened 2024-03-19
Feature presidecms 2812 async db migration priority

Pixl8RusselCole:


system/services/updateManager/AppDatabaseMigrationService.cfc
```diff
@@ -49,6 +49,22 @@ component {
 			}
 		}
 
+		if ( arguments.async ) {
+			var migrationsWithPriority = {};
+
+			for ( migration in migrations ) {
+				var priority = $runEvent(
```
DominicWatson: This will cause errors immediately if any migration does not supply a priority method. This should first do a check to see if the handler action exists

DominicWatson (changes requested):

PR #1442, opened 2024-03-25
PRESIDECMS-2819 scope variables for errorTemplate

zspitzer:
https://presidecms.atlassian.net/browse/PRESIDECMS-2819


DominicWatson:
Nice one, merging for QA. That's some old error template :imp:

PR #1443, opened 2024-03-26
PRESIDECMS-2810 newsletter send to distinct email addresses

pixl8-brayden:


system/handlers/admin/emailCenter/CustomTemplates.cfc
```diff
@@ -390,6 +392,10 @@ component extends="preside.system.base.AdminHandler" {
 		prc.filterObject = rc.filterObject = emailRecipientTypeService.getFilterObjectForRecipientType( prc.template.recipient_type ?: "" );
 		prc.anonymousOnly = !prc.filterObject.len();
 
+		if ( isEmptyString( prc.template.distinct_email ?: "" ) ) {
```
DominicWatson: We should not change the default behaviour. If that is an empty string, it should be assumed false.

system/services/email/EmailMassSendingService.cfc
```diff
@@ -137,6 +143,7 @@ component {
 				, filterParams = filterParams
 				, extraFilters = extraFilters
 				, orderBy      = idField
+				, groupBy      = groupBy
```
DominicWatson, line 148: I think we will need/want to have this groupby field in the select fields array?

DominicWatson (changes requested):

PR #1448, opened 2024-04-02
PRESIDECMS-2777 tenancy for asset is not working properly

pixl8-brayden:


system/services/assetManager/AssetManagerService.cfc
```diff
@@ -1255,8 +1258,26 @@ component displayName="AssetManager Service" {
 		return groups[ arguments.groupName ] ?: [];
 	}
 
-	public query function getAsset( required string id, array selectFields=[], boolean throwOnMissing=false ) {
-		var asset = Len( Trim( arguments.id ) ) ? _getAssetDao().selectData( id=arguments.id, selectFields=arguments.selectFields ) : QueryNew('');
+	public query function getAsset(
+		  required string  id
+		,          array   selectFields   = []
+		,          boolean throwOnMissing = false
```
DominicWatson, line 1264: It feels wrong to be adding so much code here to be manually passing around a tenant ID. My interpretation of the problem is that various background tasks are losing the tenant ID. I think the solution therefore should revolve around:

1. On bg task creation, record the current active tenants
2. On bg task run, set the current active tenants

Then the tenanting system should just take care of any filtering, etc. that is necessary. Or am I missing something?

system/services/routeHandlers/AssetRouteHandler.cfc
```diff
@@ -49,10 +55,16 @@ component implements="iRouteHandler" singleton=true presideService=true {
 		var trackDownload = buildArgs.trackDownload ?: false;
 		var trashed       = IsBoolean( buildArgs.trashed ?: "" ) && buildArgs.trashed;
 		var link          = "";
-		var assetType     = _getAssetManagerService().getAsset(
-			  id           = assetId
-			, selectFields = [ "asset_type" ]
-		).asset_type ?: "";
+		var tenantField   = Trim( _getTenancyService().getObjectTenant( "asset" ) );
```
DominicWatson, line 58: This isn't necessarily the tenant field - for that you need `getTenantFkForObject()`. Also I suggest it would make sense to cache this as a variable in this singleton (it won't change in the duration of the application lifetime). So, a `_getAssetTenancyField()` helper, that looks something like:

```cfc
if ( !StructKeyExists( variables, "_assetTenantField" ) {
    variables._assetTenantField = _getTenancyService().getTenantFkForObject( "asset" );
}
return variables._assetTenantField;
```

system/services/routeHandlers/AssetRouteHandler.cfc
```diff
@@ -93,6 +105,10 @@ component implements="iRouteHandler" singleton=true presideService=true {
 			link = event.getSiteUrl( includePath=false, includeLanguageSlug=false ) & link;
 		}
 
+		if ( Len( tenantId ) ) {
+			link &= "?tenantId=" & tenantId;
```
DominicWatson, line 109: Do not assume there is not already a query string (even though there almost certainly isn't). So the delimiter would need to be an & if detecting that there is already a qs.

system/services/assetManager/AssetQueueService.cfc
```diff
@@ -107,12 +115,14 @@ component implements="preside.system.services.assetManager.IAssetQueue" {
 		transaction {
 			var takenByOtherProcess = false;
 			var queueDao            = $getPresideObject( "asset_generation_queue" );
+			var assetTenantField    = Trim( _getTenancyService().getObjectTenant( "asset" ) );
```
DominicWatson, line 118: This is not necessarily the field, rather the _tenant_. It _usually_ is the FK field, but doesn't have to be. Use `getTenantFkForObject( "asset" )` instead. (throughout this PR)

PR #1454, opened 2024-04-08
PRESIDECMS-2826 asset download attempts to create derivative for nonexistant asset

pixl8-brayden:


system/handlers/core/AssetDownload.cfc
```diff
@@ -20,6 +20,12 @@ component {
 		var assetId           = rc.assetId      ?: "";
 		var versionId         = rc.versionId    ?: "";
 		var derivativeName    = rc.derivativeId ?: "";
+		var assetExisted      = getPresideObject( "asset" ).dataExists( id=assetId );
```
DominicWatson: This variable name is a bit off . I think we could get away without a variable here and move to service with:

```cfc
if ( !assetManagerService.assetExists( assetId ) ) {
	event.renderData( data="404 not found", type="text", statusCode=404 );
	return;
}
```

DominicWatson (changes requested):

PR #1460, opened 2024-04-10
Feature presidecms 2836 show icon in select formcontrol

Pixl8RusselCole:


system/views/formcontrols/select/index.cfm
```diff
@@ -68,6 +73,7 @@
 			<option
 				value="#selectValue#"
 				title="#label#"
+				data-iconClass="#iconClasses[i] ?: ""#"
```
DominicWatson: I think this needs removing?

DominicWatson (changes requested):

PR #1462, opened 2024-04-22
PRESIDECMS-2100 focal point picker infinite fetch loop for 404 asset

DominicWatson:
Only attempt fetching the image once and wait until we have an image loaded in order to proceed with cropping.


zspitzer:
would $fpImageClone be null, there is the img.completed which is false when there's a 404?

DominicWatson:
Yes, the point of that change was to avoid repeatedly calling `$fpImageClone = $fpImage.clone().removeAttr( "id" ).css( { position: "absolute", left: "-999em", visibility: "hidden" } ).appendTo( "body" );`. Tested working.

PR #1466, opened 2024-04-22
PRESIDECMS-2480 Apply access check prior to header setting

vivienwong:


zspitzer:
@vivienwong should the access check happen before the etag short circuit?

DominicWatson:
Thanks for this @vivienwong 

Yes, probably @zspitzer  - have added some further refactoring also to allow the interception point to do more useful things (i.e. also do https://presidecms.atlassian.net/browse/PRESIDECMS-2841 in the same work here as closely related). I've also moved the etag checkpoint.

PR #1483, opened 2024-05-22
PRESIDECMS-2873 add matrix CI jobs

zspitzer:
https://presidecms.atlassian.net/browse/PRESIDECMS-2873

### DIFF
--- a/.github/workflows/script-runner.yml
+++ b/.github/workflows/script-runner.yml
@@ -84,7 +84,7 @@ jobs:
           sudo chown -R runner:docker tests/testbox/system/stubs/
           # workaround https://luceeserver.atlassian.net/browse/LDEV-4881
           mkdir -p /home/runner/work/_actions/lucee/script-runner/main/temp/lucee/lucee-server/context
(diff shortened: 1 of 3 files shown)


DominicWatson:
I like this concept, though in its current form its not quite ready. I think:

1. We should update our testbox github image to accept jdk and lucee version arguments (or use an official Ortus action if they've now published some). Then provide the matrix variables to these actions. While the memory info is interesting, I'm not a fan of customising the output here (and it currently breaks the experience of running tests locally)

2. We need to do more shaping of the workflow, just the test running is needed to be run in a matrix so we should end up with something like:

```
[ Build dependencies ] -> [[[ Run tests ]]]  -> [ Package & Publish ]
                          JDk 11 & Lucee 5
                          JDk 11 & Lucee 6.1
                          JDk 21 & Lucee 6.1        
```

Will need to do some digging to see how much of built artifacts can be shared between jobs.

I think all the custom output and memory profile stuff is interesting, but could/should belong in a separate repo owned by Lucee if they want to use Preside as an example build to be testing things against. I imagine this could be done where the repo is pretty much just github actions and pulls down latest Preside to run against?

PR #1489, opened 2024-05-31
PRESIDECMS-2844 header visibility of environment

pixl8-brayden:


system/handlers/admin/Layout.cfc
```diff
@@ -5,9 +5,30 @@ component {
...
 	property name="adminSideBarItems"        inject="coldbox:setting:adminSideBarItems";
+	property name="environmentBannerConfig"  inject="coldbox:setting:environmentBannerConfig";
+	property name="environmentMessage"       inject="coldbox:setting:environmentMessage";
 	property name="applicationsService"      inject="applicationsService";
 	property name="i18n"                     inject="i18n";
 
+	private string function environmentBanner( event, rc, prc, args={} ) {
+		var shouldDisplay = isTrue( environmentBannerConfig.display ?: true );
+
+		if ( shouldDisplay ) {
```
sebduggan: A minor thing, but maybe clearer to return "" up front if !shouldDisplay, then everything else is not inside the conditional...

PR #1491, opened 2024-06-06
PRESIDECMS-2883 Slug does not include special character letters

teonator:


system/services/PresideSuperClass.cfc
```diff
@@ -925,10 +925,15 @@ component displayName="Preside Super Class" {
 	public string function $slugify( required str, numeric maxLength=0, allow="", preserveCase=false ) {
 		var slug = Trim( arguments.str );
 
+		var specialChars = "#chr(225)#,#chr(224)#,#chr(226)#,#chr(228)#,#chr(227)#,#chr(229)#,#chr(233)#,#chr(232)#,#chr(234)#,#chr(235)#,#chr(237)#,#chr(236)#,#chr(238)#,#chr(239)#,#chr(243)#,#chr(242)#,#chr(244)#,#chr(246)#,#chr(245)#,#chr(248)#,#chr(250)#,#chr(249)#,#chr(251)#,#chr(252)#,#chr(241)#,#chr(231)#,#chr(223)#,#chr(193)#,#chr(192)#,#chr(194)#,#chr(196)#,#chr(195)#,#chr(197)#,#chr(201)#,#chr(200)#,#chr(202)#,#chr(203)#,#chr(205)#,#chr(204)#,#chr(206)#,#chr(207)#,#chr(211)#,#chr(210)#,#chr(212)#,#chr(214)#,#chr(213)#,#chr(216)#,#chr(218)#,#chr(217)#,#chr(219)#,#chr(220)#,#chr(209)#,#chr(199)#";
+		var asciiChars   = "a,A,a,ae,A,a,e,E,e,E,e,i,I,i,I,o,O,o,O,o,o,u,U,ue,U,n,c,ss,A,A,A,A,A,A,A,A,A,A,E,E,E,E,E,E,I,I,I,I,I,I,O,O,O,O,O,O,O,O,O,O,O,U,U,U,U,U,U,N,C";
+
+		slug = ReplaceList( slug, specialChars, asciiChars );
```
zspitzer, line 931: might be clearer to use a https://docs.lucee.org/reference/functions/replace.html with a struct, that way the above dual lists would be easier to follow `{chr(255): "a",.... }` also way better performance as lists require string processing each time while the struct is pre-parsed

PR #1503, opened 2024-06-28
Feature pressynced 2626 500 error on send legacy email

mookeshNanda:
Check template record exists first before trigger updateLastSentDate


DominicWatson:
Fab, thanks @mookeshNanda - getting this released asap

PR #961, opened 2021-03-22
Fixed the issue for Text Area validation error for new line

cfmitrah:
https://presidecms.atlassian.net/browse/PRESIDECMS-1752

PR #1509, opened 2024-07-17
Feature presidecms 2771 incorrect textarea character count in admin

jamesallenuk:
This change reverts changes made in PR - [#961] (Mar 31, 2021) due to issues it was causing with incorrect character count adding new lines to the count.
Changes were reverted and logic has been added to normalise the text content to ensure \r\n characters are changed to \n before counts are made.
This same logic has also added to the FormsService to ensure received text content is normalised in the same way before being stored.

### DIFF
--- a/system/assets/js/admin/lib/plugins/jquery.validate.js
+++ b/system/assets/js/admin/lib/plugins/jquery.validate.js
@@ -1385,9 +1385,7 @@ $.extend( $.validator, {
 
 		// http://jqueryvalidation.org/maxlength-method/
 		maxlength: function( value, element, param ) {
-			value = value.replace(/\r\n/g, '\n');
-
-			var length = $.isArray( value ) ? value.length : this.getLength( value, element );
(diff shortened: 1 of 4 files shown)


system/assets/js/admin/lib/plugins/jquery.validate.js
```diff
@@ -1385,11 +1385,9 @@ $.extend( $.validator, {
 
 		// http://jqueryvalidation.org/maxlength-method/
 		maxlength: function( value, element, param ) {
-			var newLines = value.match(/\r\n|\n|\r/g), addition = 0;
-			if (newLines != null) {
-				addition = newLines.length;
-			}
-			var length = $.isArray( value ) ? value.length + addition : this.getLength( value, element ) + addition;
+			value = value.replace(/\r\n/g, '\n');
```
sebduggan: As per the line below, the value might be an array. We should only do this to the value if it's not an array...

PR #1518, opened 2024-08-09
PRESIDECMS-2920 text property match expression with space not returning expected result

choontat-pixl8:
Use `ListItemTrim` only if the operator is `oneof` or `noneof`


DominicWatson:
Nice one, thanks @choontat-pixl8 . Merged into release for QA

PR #1532, opened 2024-09-12
Feature presidecms 2924 500 error when using selectunion with saved filters

Pixl8RusselCole:


system/services/presideObjects/PresideObjectService.cfc
```diff
@@ -459,11 +461,74 @@ component displayName="Preside Object Service" {
...
+		for ( var matched in matches ) {
+			if ( !Len( matched.match[1] ) ) {
+				continue;
+			}
+			
+			var decoded = ToString( ToBinary( ReReplace( matched.match[1], obfsPattern, "\1" ) ) );
+			ArrayAppend( obfuscations, {
+				  encoded      = matched.match[1]
+				, decoded      = decoded
+				, obfuscations = _getObfuscationsInSql( decoded, arguments.cnt+1 )
```
DominicWatson: can't see `arguments.cnt` being defined anwhere. Also, let's not use `cnt` - `recursionCount` is clearer + doesn't look rude :p

DominicWatson (changes requested):

PR #1536, opened 2024-09-27
PRESIDECMS-2944 outputView & outputViewlet

DominicWatson:
Adds more options for developers with regards rendering views and viewlets. See https://community.preside.org/t/an-interesting-performance-problem-and-proposed-new-helpers-to-alleviate/178/2

### DIFF
--- a/system/coldboxModifications/Controller.cfc
+++ b/system/coldboxModifications/Controller.cfc
@@ -104,6 +104,7 @@ component extends="coldbox.system.web.Controller" {
 		,          string  cacheLastAccessTimeout = ""
 		,          string  cacheSuffix            = ""
 		,          string  cacheProvider          = "template"
+		,          boolean throwOnMissing         = true
 	) {
 		if ( arguments.delayed && _getDelayedViewletRendererService().isDelayableContext() ) {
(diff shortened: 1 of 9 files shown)


system/coldboxModifications/Controller.cfc
```diff
@@ -133,37 +134,126 @@ component extends="coldbox.system.web.Controller" {
...
-				, eventArguments = { args = arguments.args }
+				, eventArguments = { args = viewletArgs }
 			);
+
+			if ( IsNull( local.result ) ) {
+				view            = requestContext.getViewletView();
+				deferredViewlet = requestContext.getDeferredViewlet();
+				viewletArgs     = requestContext.getViewletArgs( viewletArgs );
+			}
+			requestContext.popViewletContext();
```
Pixl8RusselCole, line 166: Can we move this higher up and return early? (ie before if ( IsNull( local.result ) )

if ( !IsNull( local.result ) ) {

DominicWatson: No, because `requestContext.getViewletView()`, etc. all depend on the viewlet context being the current viewlet. Can/must pop it after that.

PR #1541, opened 2024-10-09
PRESIDECMS-2949 Existing oneToManyConfigurator field value does not show when editing

teonator:


system/handlers/formcontrols/OneToManyConfigurator.cfc
```diff
@@ -10,6 +10,7 @@ component {
 		var sourceIdField  = presideObjectService.getIdField( args.sourceObject );
 		var sortOrderField = presideObjectService.getObjectAttribute( targetObject, "datamanagerSortField", "sort_order" );
 		var hasSortOrder   = StructKeyExists( presideObjectService.getObjectProperties( targetObject ), sortOrderField );
+		var isEdit         = isEmptyString( args.savedData[ "datecreated" ] ?: "" );
```
DominicWatson, line 13: This naming seems the wrong way around? `isEdit` should be `= !isEmptyString( ... )`? and then the logic below would be `if ( isEdit || isTrue( sourceProperty.cloneable...`?

I wonder should it all be simpler and just use `if ( !StructIsEmpty( args.savedData ) )`? i.e. if we have saved data, whether cloning or editing, then we should prepopulate?

I'm curious that this would not have been picked up previously re the bug in general. Wonder if @sebduggan has any thoughts here?

DominicWatson (changes requested):

system/handlers/formcontrols/OneToManyConfigurator.cfc
sebduggan: The issue was only introduced with the previous bugfix from last month...

teonator: I am afraid we cannot use `!StructIsEmpty(args.savedData)` because the cloned `args.savedData` is not empty, but it is missing datecreated and datemodified.

PR #1554, opened 2024-12-16
PRESIDECMS-2906 Client side validation defensive when validator didnt have client side defined

teonator:


system/views/formcontrols/layouts/form.cfm
```diff
@@ -53,13 +53,18 @@
 	</cfif>
 
 	<cfif Len( Trim( formId ) ) and Len( Trim( validationJs ))>
+		<cfset event.include( assetId="/js/frontend/formbuilder/" )>
```
DominicWatson, line 56: I'm a little sceptical of this. We're making a dependency on form builder frontend features whereas this layout is for the general preside forms system and is used in both admin and frontend non-formbuilder forms. I feel this logic could/should go somewhere else perhaps?

DominicWatson (changes requested):

PR #1559, opened 2024-12-20
PRESIDECMS-2956 Formbuilder form export error when response exceed 32767 characters

teonator:


system/services/formbuilder/FormBuilderService.cfc
```diff
@@ -1639,10 +1640,18 @@ component {
 					var mappedColumns = itemColumnMap[ item.id ];
 
 					for( var i=1; i<=mappedColumns.len(); i++ ) {
+						++column;
```
jamesallenuk: While no logical difference I'd go for `column++` here and it would then be consistent with how row is incremented above.

teonator: Updated increment consistency.

system/services/formbuilder/FormBuilderService.cfc
```diff
@@ -1664,6 +1680,34 @@ component {
 			}
 		}
 
+		if ( !StructIsEmpty( extraData ) ) {
+			if ( canInfo ) {
+				logger.info( "Processing additional data..." );
+			}
+
+			$helpers.dumpLog(extraData=extraData);
```
jamesallenuk: Should this be left in?

PR #1561, opened 2025-01-17
PRESIDECMS-2794 Add a way to verify authentication timeout via cookie instead of IP address

choontat-pixl8:


DominicWatson:
I like the general gist of this. However, `2faAuthMethod` is not an appropriate name for this setting (auth method suggests the actual method of authentication, rather than how we attempt to maintain auth over the configured period of time) and the language around the configuration is problematic.

I'd be tempted to suggest we just don't make this configurable at all. But if we really feel that we must, then naming of variables etc. should be something like `2faPersistenceMethod` - explaining that to end-users in the settings enum is a little tricky (and what is there now is also not enough really).

Did you consider not having it configurable?

system/services/admin/LoginService.cfc
```diff
@@ -677,20 +678,23 @@ component displayName="Admin login service" {
...
+
+			_getCookieService().setVar( name=_getTwoFactorAuthCookiePersistKey(), value=authMethodCookieValue );
+
 			var loginRecordDao = $getPresideObject( "security_user_two_factor_login_record" );
-			var updated = loginRecordDao.updateData( filter={
-				  security_user = userId
-				, ip_address    = arguments.ipAddress
-				, user_agent    = arguments.userAgent
-			}, data={ logged_in_date=Now() } );
+			var updated        = loginRecordDao.updateData(
```
DominicWatson: I think this `updateData` logic needs to go. Previously, IP + user agent was a unique combination. We are now shifting that to the cookie being the unique part. We know that this is a new cookie value so we can just go ahead and insert a new record and not attempt to update records that happen to match the user agent.

system/preside-objects/admin/security/security_user_two_factor_login_record.cfc
```diff
@@ -13,5 +13,6 @@ component extends="preside.system.base.SystemPresideObject" displayName="User 2F
 
 	property name="ip_address"     type="string"   dbtype="varchar" maxLength="50"  required=true uniqueindexes="user_machine|2";
```
DominicWatson: I believe we can and should get rid of the IP + user agent fields here. At the very least, the unique index needs to go.

system/preside-objects/admin/security/security_user_two_factor_login_record.cfc
```diff
@@ -13,5 +13,6 @@ component extends="preside.system.base.SystemPresideObject" displayName="User 2F
...
 	property name="user_agent"     type="string"   dbtype="varchar" maxLength="255" required=true uniqueindexes="user_machine|3";
+	property name="cookie"         type="string"   dbtype="varchar" maxLength="255";
```
DominicWatson: The cookie value should have a unique index defined.

system/services/admin/LoginService.cfc
```diff
@@ -525,9 +526,9 @@ component displayName="Admin login service" {
 			  selectFields = [ "logged_in_date" ]
 			, filter       = {
 				  security_user = getLoggedInUserId()
-				, ip_address    = arguments.ipAddress
```
DominicWatson, line 528: The user agent should also go from the filter here.

system/services/admin/LoginService.cfc
```diff
@@ -677,20 +678,23 @@ component displayName="Admin login service" {
...
 
 			if ( !updated ) {
-				loginRecordDao.insertData({
-					  security_user  = userId
-					, ip_address     = arguments.ipAddress
-					, user_agent     = arguments.userAgent
-					, logged_in_date = Now()
-				});
+				loginRecordData.security_user = userId;
+				loginRecordData.user_agent    = arguments.userAgent;
```
DominicWatson: Remove the user agent and IP here if removing from the object (see prior feedback)

DominicWatson (changes requested):

system/services/admin/LoginService.cfc
```diff
@@ -677,20 +678,23 @@ component displayName="Admin login service" {
 				two_step_auth_key_in_use = true
 			} );
 
+			var authMethodCookieValue = CreateUUID();
```
DominicWatson: just `cookieValue` is fine here. `authMethod` adds nothing and is misleading.

DominicWatson (changes requested):

system/preside-objects/admin/security/security_user_two_factor_login_record.cfc
```diff
@@ -11,7 +11,6 @@
 component extends="preside.system.base.SystemPresideObject" displayName="User 2FA Login record" {
 	property name="security_user" relationship="many-to-one" relatedTo="security_user" required=true uniqueindexes="user_machine|1";
 
-	property name="ip_address"     type="string"   dbtype="varchar" maxLength="50"  required=true uniqueindexes="user_machine|2";
-	property name="user_agent"     type="string"   dbtype="varchar" maxLength="255" required=true uniqueindexes="user_machine|3";
-	property name="logged_in_date" type="datetime" dbtype="datetime"                required=false;
+	property name="cookie"         type="string"   dbtype="varchar"  maxLength="255" required=true uniqueindexes="user_machine|2";
```
DominicWatson: We cannot add a required field that is part of a unique index into an existing table with data - this will fail migration.

An alternative here would be to just use the login record ID as the token value. Then remove the "user_machine" unique index entirely.

system/services/admin/LoginService.cfc
```diff
@@ -677,21 +671,15 @@ component displayName="Admin login service" {
...
-
-			if ( !updated ) {
-				loginRecordDao.insertData({
-					  security_user  = userId
-					, ip_address     = arguments.ipAddress
-					, user_agent     = arguments.userAgent
-					, logged_in_date = Now()
-				});
-			}
+			var cookieValue = CreateUUID();
```
DominicWatson: as per note above, this could be refactored:

```cfc
var loginRecordId = $getPresideObject( "security_user_two_factor_login_record" ).insertData( {
	  security_user  = userId
	, logged_in_date = Now()
} );
_getCookieService().setVar( name=_getTwoFactorAuthCookiePersistKey(), value=loginRecordId );
```

DominicWatson (changes requested):

system/preside-objects/admin/security/security_user_two_factor_login_record.cfc
```diff
@@ -9,9 +9,7 @@
  * @feature   twoFactorAuthentication
  */
 component extends="preside.system.base.SystemPresideObject" displayName="User 2FA Login record" {
-	property name="security_user" relationship="many-to-one" relatedTo="security_user" required=true uniqueindexes="user_machine|1";
+	property name="security_user" relationship="many-to-one" relatedTo="security_user" required=true uniqueindexes="user_machine";
```
DominicWatson: The unique index has to go. A single user can be logged in to multiple devices, each with its own cookie so there could be multiple entries with the same user.

DominicWatson (changes requested):

DominicWatson:
Fab, thanks CT. Merging for QA

PR #1571, opened 2025-02-19
Feature presidecms 3018 object default filter not working

mookeshNanda:
Used args.savedFilters instead as this contain the appended object defaultFilter


DominicWatson:
Thanks @mookeshNanda. Will get this released asap

PR #1573, opened 2025-02-26
Feature presidecms 3020 email layout config item missing i18n and label

s-ven-315:


system/handlers/renderers/labels/email_layout_config_item.cfc
```diff
@@ -0,0 +1,24 @@
...
+	}
+
+	private string function _renderLabel( event, rc, prc ) {
+		var template_name   = arguments.template_name  ?: "";
+		var blueprint_name  = arguments.blueprint_name ?: "";
+		var item            = arguments.item           ?: "";
+
+		var label = [ item ];
+		if( Len( template_name ) ){
+			ArrayPrepend( label, "Template: #template_name#" );
```
Pixl8RusselCole: I think we need i18n for Template: and Blueprint:   please

PR #1579, opened 2025-03-21
PRESIDECMS-3033 performance fix for some mysql queries

DominicWatson:
Use subquery when doing both an order AND a limit. This is a known workaround for seemingly random issues with some
queries where this combination leads to very slow queries.


DominicWatson:
This approach was bogus and the tests saved us!

PR #1588, opened 2025-04-10
Feature presidecms 3053 draft page cannot be published without making changes

s-ven-315:


system/services/siteTree/SiteTreeService.cfc
```diff
@@ -838,7 +838,8 @@ component {
 
 			versionNumber = _getPresideObjectService().getNextVersionNumber();
 
-			var pageDataHasChanged     = _getVersioningService().dataHasChanged( objectName="page", recordId=arguments.id, newData=arguments );
+			var isPublishingDraftPage  = !arguments.isDraft && $helpers.isTrue( existingPage._version_is_draft );
+			var pageDataHasChanged     = _getVersioningService().dataHasChanged( objectName="page", recordId=arguments.id, newData=arguments ) || isPublishingDraftPage;
```
DominicWatson, line 842: This all looks logically great. Just a quick note that is best to perform cheap logic first when stringing expressions. i.e.:

```
if ( somethingCheap || somethingMoreExpensive )
```

In this case:

```cfc
var pageDataHasChanged = isPublishingDraftPage || _getVersioningService().dataHasChanged( objectName="page", recordId=arguments.id, newData=arguments );
```

This way, if we're publishing a draft page, we never have to run the more expensive logic to check whether or not data has changed.

DominicWatson: (I'm merging this and making that change). thanks for the PR!

PR #1589, opened 2025-04-10
PRESIDECMS-3047 Save pages completed by the user for a multi page form

teonator:


system/handlers/renderers/content/FormBuilderSubmissionFormPage.cfc
```diff
@@ -0,0 +1,27 @@
...
+		var renderedPages = [];
+		var pages         = formBuilderService.getPages( formId=formId );
+
+		for ( var page in pages ) {
+			if ( ListContains( args.data, page.id ) ) {
+				ArrayAppend( renderedPages, page.configuration.label );
+			}
+		}
+
+		return '<ul><li>#ArrayToList( renderedPages, "</li><li>" )#</li></ul>';
```
DominicWatson: I think this should be as follows to avoid return empty html elements unnecessarily:

```cfc
if ( ArrayLen( renderedPages ) ) {
    return '<ul><li>#ArrayToList( renderedPages, "</li><li>" )#</li></ul>';
}
return "";
```

DominicWatson (changes requested):
Looking great, just one little comment on a renderer

PR #1596, opened 2025-04-17
PRESIDECMS-2981 Unclear 200 character limit on front end of question values field

teonator:


system/i18n/formbuilder.properties
```diff
@@ -95,4 +95,6 @@ cloneForm.page.title=Clone form
 cloneForm.save=Save
 cloned.success.message=Form has been cloned successfully.
 
-task.form.delete.title=Delete form and all its data
\ No newline at end of file
+task.form.delete.title=Delete form and all its data
+
+validation.responseMaxLength.default=One of the line item must be no more than {2} characters.
```
DominicWatson: > Each line item must be no more than {2} characters.

DominicWatson (changes requested):

PR #1602, opened 2025-05-05
PRESIDECMS-3070 detect incoming admin site wasnt set when didnt have site navigate permissions

pixl8-brayden:


system/coldboxModifications/services/RoutingService.cfc
```diff
@@ -104,6 +104,10 @@ component extends="coldbox.system.web.services.RoutingService" accessors=true {
 
 		_announceInterception( "onPresideDetectIncomingSite", interceptData );
 
+		if ( StructIsEmpty( interceptData.site ) ) {
+			interceptData.site = siteService.matchSite( domain=domain, path=pathInfo );
+		}
+
```
jamesallenuk: Shouldn't this be before the interception is announced so that the site is populated if missing and included in the interceptData sent out?

PR #1610, opened 2025-05-28
PRESIDECMS-3084 Formbuilder web request context

teonator:


system/handlers/rules/contexts/WebRequest.cfc
```diff
@@ -0,0 +1,17 @@
+component {
```
DominicWatson: I think this logic change should probably go in `/handlers/rules/contexts/FormbuilderSubmission.cfc` (that already exists). Feels out of place here.

DominicWatson (changes requested):

PR #1614, opened 2025-06-12
PRESIDECMS-3094 add twitter meta

seakchiew:
Add support for twitter card meta to allow image to be display in default summary style on twitter

### DIFF
--- a/system/config/Config.cfc
+++ b/system/config/Config.cfc
@@ -1324,7 +1324,7 @@ component {
 		derivatives.openGraphImage = {
 			  permissions = "inherit"
 			, autoQueue   = [ "image" ]
-			, transformations = [ { method="shrinkToFit", args={ width=400, height=400 } } ]
+			, transformations = [ { method="shrinkToFit", args={ width=1200, height=630 } } ]
 		};
 
 		return derivatives;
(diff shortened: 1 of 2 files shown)


system/config/Config.cfc
```diff
@@ -1324,7 +1324,7 @@ component {
 		derivatives.openGraphImage = {
 			  permissions = "inherit"
 			, autoQueue   = [ "image" ]
-			, transformations = [ { method="shrinkToFit", args={ width=400, height=400 } } ]
+			, transformations = [ { method="shrinkToFit", args={ width=1200, height=630 } } ]
```
DominicWatson: Ideally this is dealt with in a separate ticket. That can be communicated. Also, based on the guidelines, `shrinkToFit 1200x1200` seems more appropriate to work with different orientations? The dimensions wouldn't always be exactly what is called for, but should be good enough.

DominicWatson (changes requested):

system/config/Config.cfc
seakchiew: Hi @DominicWatson  i have changed the derivative to 1200 x 1200.

PR #1618, opened 2025-06-18
PRESIDECMS-3100 500 error in star rating formcontrol when allowhalfstars is missing

teonator:


system/handlers/rules/fieldtypes/formbuilderQuestionStarRatingValue.cfc
```diff
@@ -4,25 +4,26 @@
...
-		rc.delete( "value" );
-		var questionId = config.question ?: ( rc.question ?: "" );
+		StructDelete( rc, "value" );
+
+		var questionId  = arguments.config.question ?: ( rc.question ?: "" );
 		var theQuestion = formBuilderService.getQuestion( questionId );
 
 		if ( len( theQuestion ) ) {
-			var questionConfig = DeserializeJson( theQuestion.item_type_config );
+			var questionConfig = DeserializeJSON( theQuestion.item_type_config );
```
DominicWatson: `DeserializeJson()` preferred here. We treat common acronyms as words with our camel case style. So `Id`, vs `ID`, etc.

PR #1625, opened 2025-07-02
PRESIDECMS-3110 no able to include asset resource in form

pixl8-brayden:


DominicWatson:
Nice one, been meaning to do this for some time and I think there may be an existing story floating around somewhere!

Inline with other form approaches though, we should add support for this attribute on `tab`, `fieldset` and `field` elements also. This makes it easy for developers to extend forms and add js/css to elements that may or may not always be available.

PR #1663, opened 2025-09-17
PRESIDECMS-3127 fix instance prepare filter active type not working

pixl8-brayden:


system/services/webflow/WebflowInstanceService.cfc
```diff
@@ -468,7 +468,11 @@ component {
 			} );
 		} else {
 			var timeoutOperator = arguments.type == "active" ? ">=" : "<";
-			var webflowConfig   = $getPresideObject( "webflow_configuration" ).selectData( id=arguments.webflowId );
+			var webflowConfig   = $getPresideObject( "webflow_configuration" ).selectData(
```
Pixl8RusselCole, line 471: Any reason not to use a returnType  (singleValue) here?

PR #1672, opened 2025-09-25
Feature presidecms 3151 saved filters should namespace their parameters to avoid collisions

jamesallenuk:


system/services/presideObjects/PresideObjectService.cfc
```diff
@@ -3919,9 +3922,71 @@ component displayName="Preside Object Service" {
...
+						break;
+					}
+				}
+			}
+
+			if ( hasCollision ) {
+				var sqlParts = [];
+				for ( var key in extraFilter.filter ) {
+					if ( StructKeyExists( originalFilterParams, key ) ) {
+						var paramName = "sf_" & ReReplace( key, "[\.\$]", "__", "all" );
```
sebduggan: Any namespacing needs to be unique for each extraFilter, as just "sf_" prefix could still cause collisions between multiple extraFilters that reference the same property (and "sf_" is inaccurate, as it's not just saved filters). Maybe a hash of key and values? And set the value once in each extraFilter iteration.

system/services/presideObjects/PresideObjectService.cfc
```diff
@@ -3919,9 +3922,71 @@ component displayName="Preside Object Service" {
...
-				result.filterParams.append( extraFilter.filter );
+				for ( var key in extraFilter.filter ) {
+					if ( StructKeyExists( originalFilterParams, key ) ) {
+						hasCollision = true;
+						break;
+					}
+				}
+			}
+
+			if ( hasCollision ) {
```
sebduggan: Should this all exist in the above "if isStruct()" block?

PR #1760, opened 2026-04-11
Add chunked file upload support to asset manager PRESIDECMS-3260

alexskinner:
Splits large files into 10MB chunks client-side and reassembles them server-side, bypassing proxy/CDN request size limits (e.g. Cloudflare's 100MB limit) while still enforcing Preside's own file size restrictions. The chunking threshold is configurable via settings.assetManager.chunkingThreshold.

New files:
- ChunkedUploadService.cfc — chunk storage, assembly, stale session cleanup
- ChunkedUpload.cfc handler — uploadChunk, finalize, uploadNewVersionPage, finalizeNewVersion actions
- uploadNewVersionPage.cfm — simplified single-file version upload UI
- chunkedUploadInterface.js — client-side chunking, Dropzone integration, folder restriction re-validation, compiled Sticker bundle

AssetManager changes:
- uploadAssets, assetPickerUploader, getUploadRestrictions now always include chunked upload URLs and pass the full system extension list to the client when no folder restriction is set
- New getUploadRestrictions AJAX action for re-validating queued files when folder selection changes
- Upload new version button now links to chunkedUpload.uploadNewVersionPage replacing the old hidden file input approach

Bugs discovered and fixed during development (unrelated to chunking):
- uploadAssets: event.includeData was inside the folder.recordCount condition, so chunk URLs and maxFileSize were never sent when uploading to the root folder (no folder selected)
- uploadAssets: prc.pageIcon was "picture" instead of "picture-o", causing a missing icon in the admin breadcrumb
- uploadAssets / assetPickerUploader: unknown file types (e.g. .dmg) were accepted client-side when no folder restriction was set because allowedExtensions was never populated from the system type list, causing a server-side exception instead of a validation message

i18n: added chunked upload error and success keys; all user-facing messages use plain English with no technical terminology

---

> [!NOTE]
> **Medium Risk**
> Introduces a new upload path that streams and reassembles file chunks server-side, touching core asset upload/versioning flows and temporary file handling. Risk is mainly around incomplete chunk sessions, cleanup behavior, and parity with existing per-folder restriction validation.
> 
> **Overview**
> Adds **client/server chunked upload support** to the admin Asset Manager so large files can be uploaded in multiple requests (configurable via `settings.assetManager.chunkingThreshold`, default 10MB) and then finalized into a single stored asset.
> 
> Updates the upload pages to always provide `allowedExtensions` (falling back to the full system type list), plus new chunk endpoints and an AJAX `getUploadRestrictions` call to re-validate queued files when the target folder changes.
> 
> Reworks “upload new version” to use a dedicated chunked-upload page and server finalize action, replacing the previous hidden file-input flow, and adds new i18n strings for chunked upload success/failure messages.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 076c5e16103bba5af8d957c1a78e4a06bc40a0af. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


system/handlers/admin/ChunkedUpload.cfc
```diff
@@ -0,0 +1,238 @@
...
+					  success = false
+					, message = translateResource( "cms:assetmanager.chunked.upload.error.no.chunk.data" )
+				} );
+				return;
+			}
+
+			var saved = chunkedUploadService.saveChunk(
+				  uuid        = uuid
+				, chunkNumber = chunkNumber
+				, chunkData   = FileReadBinary( chunkPath )
```
DominicWatson: Is there anyway to avoid reading the file - can we just be copying file paths instead? (makes for leaner memory usage if we can).

system/handlers/admin/ChunkedUpload.cfc
```diff
@@ -0,0 +1,238 @@
...
+			, totalChunks = totalChunks
+		);
+
+		if ( !assembled.success ) {
+			event.renderData( type="json", data={
+				  success = false
+				, message = translateResource( "cms:assetmanager.chunked.upload.error.assemble" )
+			} );
+			return;
+		}
```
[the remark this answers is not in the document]
DominicWatson: Valid ^^

system/handlers/admin/ChunkedUpload.cfc
```diff
@@ -0,0 +1,238 @@
...
+			} );
+		} catch ( any e ) {
+			logError( e );
+			event.renderData( type="json", data={
+				  success = false
+				, message = translateResource( "cms:assetmanager.chunked.upload.error.chunk", { data=[ e.message ] } )
+			} );
+		} finally {
+			// Clean up the temp directory after addAssetVersion is done
+			var tempDir = GetTempDirectory() & "preside_chunked_" & uuid & "/";
```
DominicWatson: This should be moved to a cleanupTmpDir() method (or similar name) in the chunkingService so that all the temp directory logic is in there.

system/services/chunkedUpload/ChunkedUploadService.cfc
```diff
@@ -0,0 +1,191 @@
...
+	 * The temp dir is created on first chunk. No separate session step required.
+	 *
+	 * @uuid        Unique upload identifier generated client-side
+	 * @chunkNumber 1-based chunk index
+	 * @chunkData   Binary chunk data
+	 */
+	public boolean function saveChunk(
+		  required string  uuid
+		, required numeric chunkNumber
+		, required any     chunkData
```
DominicWatson: As previously stated, this could be passed as a filepath and we can then just move the file rather than reading the 10mb binary into memory and then writing it back to file.

system/services/chunkedUpload/ChunkedUploadService.cfc
```diff
@@ -0,0 +1,191 @@
...
+	/**
+	 * Returns a canonical, symlink-resolved temp path for the given UUID.
+	 * Uses Java getCanonicalPath() to ensure consistency on macOS where
+	 * /tmp is a symlink to /private/tmp.
+	 */
+	private string function _getTempDir( required string uuid ) {
+		var canonicalTempBase = createObject( "java", "java.io.File" ).init( GetTempDirectory() ).getCanonicalPath();
+		if ( Right( canonicalTempBase, 1 ) != "/" ) { canonicalTempBase &= "/"; }
+		return canonicalTempBase & "preside_chunked_" & arguments.uuid & "/";
+	}
```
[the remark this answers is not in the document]
DominicWatson: This ^^. Also the whole java getCanonicalPath should not be necessary. Just something like:

```cfc
if ( !isValid( "uuid", arguments.uuid ) { // note, isValid( "uuid", uuid ) only returns true for cfml uuids, might need a helper function + some regex
    // throw...
}
return ListAppend( ReReplace( GetTempDirectory(), "/$", "" ), "preside_chunked_#arguments.uuid#", "/" );
```

system/services/chunkedUpload/ChunkedUploadService.cfc
```diff
@@ -0,0 +1,191 @@
...
+
+	/**
+	 * Clean up abandoned upload directories older than 2 hours.
+	 * Called before each assembly to avoid temp dir accumulation.
+	 */
+	public void function cleanupStaleSessions() {
```
DominicWatson: This should not be necessary. Lucee does its own tmp directory cleanup

DominicWatson (changes requested):
Looking great. Few bits of review. Particularly important is the tmp directory feedback.

PR #1795, opened 2026-06-16
PRESIDECMS-3303 locale number formatting helpers improvement

pixl8-brayden:


system/services/l10n/NumberFormatService.cfc
```diff
@@ -0,0 +1,40 @@
...
+component {
+
+	public any function init() {
+		return this;
+	}
+
+	public string function getNumberFormatMask( string locale="" ) {
+		var resolvedLocale = Len( Trim( arguments.locale ) ) ? Trim( arguments.locale ) : _getSiteLocale();
+
+		switch ( resolvedLocale ) {
```
Pixl8RusselCole, line 14: Just checking that this is supposed to return "," in all cases

PR #1811, opened 2026-07-29
Encode value rendering in admin

pixl8-brayden:


DominicWatson:
I'm not sure about this one. It would make immediate changes everywhere that may not be wanting html encoding and could therefore break applications. That replaceList logic that is removed appears to have been doing the opposite also. Not sure.

PR #1814, opened 2026-08-10
PRESIDECMS-3323 inherited rule handler component metadata is ignored

choontat-pixl8:
Get the extended component metadata, including
- component properties (e.g. `@expressionContexts`)
- functions, function parameters (e.g. `required string type`) and their properties (e.g. `@type.fieldType`)

Also updates the expression test service to pass `Duplicate( meta.functions[1] )` to match the new changes


system/services/rulesEngine/RulesEngineExpressionReaderService.cfc
```diff
@@ -114,37 +114,27 @@ component displayName="RulesEngine Expression Reader Service" {
...
-					, filterHandler         = filterObjects.len() ? "rules.expressions.#baseId#.prepareFilters" : ""
-					, labelHandler          = "rules.expressions.#baseId#.getLabel"
-					, textHandler           = "rules.expressions.#baseId#.getText"
-					, expressionHandlerArgs = {}
-					, filterHandlerArgs     = {}
-					, labelHandlerArgs      = {}
-					, textHandlerArgs       = {}
-				};
-
-			} else if ( func.name == "prepareFilters" ) {
```
Pixl8RusselCole, line 140: Was it your intention to remove prepareFilters completely?

tests/.gitignore
```diff
@@ -3,3 +3,4 @@ web.config
 testbox
 testresults*.html
 .cfconfig.json
+server-presidetests.json
```
Pixl8RusselCole: Not a biggy but I don't think this should be in this ticket.

system/services/rulesEngine/RulesEngineExpressionReaderService.cfc
choontat-pixl8: The original `prepareFilters` check is to ensure that the `filterObjects` function property is set in the expression after the `evaluateExpression` function is processed.

The current `_getExtendedFunctionsMeta` function returns a struct of function meta, for example
```
{
      prepareFilters     = { objects="user,contact", parameters=[] }
    , evaluateExpression = { parameters=[] }
}
```

So we can now do 
```
var filterObjects = ListToArray( functions.prepareFilters.objects ?: "" );
...
expressions[ baseId ] = {
      filterObjects = filterObjects
    , filterHandler = filterObjects.len() ? "rules.expressions.#baseId#.prepareFilters" : ""
    ...
};
```

tests/.gitignore
choontat-pixl8: This line is now removed alongside with the `server-presidetests.json` file

